PR #13, opened 2020-04-29
Provide CRUD ops for project profiles

### LINKED ISSUE #14: Users should be able to read project profiles from the registry
Implement the ability to fetch all projects or a single project by ID.

- [ ] OpenAPI doc
- [ ] Implement endpoints
- [ ] Automated tests

jleach:
This PR provides basic CRUD operations for a project "Profile". This includes initial work on the db schema, API, and data modelling pattern. I'm trying to not use an ORM or query builder; just using simple SQL queries with the `pg` driver. I'm also working to make the project generally as testable as possible based on previous learning with the signing service and repomountie. ATM the API does not have any authentication or authorization; I'm waiting to get some calls implemented before I secure them.

Fixes #14


jleach:
@patricksimonian @ShellyXueHan In addition to a general PR looking for feedback on design. I'm going with a simple Express API in TypeScript; the data model is done via a DataManager that just abstracts away some basic SQL queries - no ORM no query builders. Just back to basics.

api/src/libs/profile.ts
```diff
@@ -18,6 +18,40 @@
...
+  } catch (err) {
+    const message = 'Unable fetch all project profiles';
+    logger.error(`${message}, err = ${err.message}`);
+
+    throw errorWithCode(message, 500);
+  }
+};
+
+export const fetchProjectProfile = async (req: Request, res: Response): Promise<void> => {
+  const dm = new DataManager(config);
```
ShellyXueHan, line 43: use a shared dm instance?

api/src/config/index.ts
```diff
@@ -42,39 +42,29 @@ if (env === 'development') {
...
-    key: process.env.SESSION_SECRET,
-  },
 });
 
 // load other properties from file.
 nconf
   .argv()
   .env()
-  .file({ file: path.join(__dirname, `${env}.json`) });
+  .file({ file: path.join(__dirname, `${fileName}`) });
```
patricksimonian, line 62: does this config file live relative to this directory? Is there a need for the config to be different in different environments (production, test etc)? If so, does this allow a configmap to be mounted

jleach: I mounted it as a configmap now. This way we don't need to do a new build to change config params.

api/src/libs/profile.ts
jleach: I'm experimenting a bit with connection `Pools` to see if they're shared automatically by pg.  Haven't quite settled on the final pattern.

PR #23, opened 2020-05-08
Add functionality for CRUD on project namespaces

### LINKED ISSUE #7: The API should have the concept of namespaces
The API should have the concept of namespaces that relate back to a profile (project). This will be used to track what namespaces belong to project.

- [ ] Create profile endpoints in OpenAPI doc;
- [ ] Create data model in Knex;
- [ ] Create API endpionts.
  - [ ] Create
  - [ ] Read
  - [ ]  Update
  - [ ] Delete

### LINKED ISSUE #21: Projects should show what cluster(s) they run on
We're going to have 6 OCP4 clusters, one of which will be DR. What cluster(s) the namespaces exist on should be reflected in the registry.

jleach:
Add functionality for CRUD on project namespaces. This includes what cluster the namespace is provisioned on. Other changes include fixing minor issues with the db schema. I also added vscode config to allow it to debug in Docker containers for improved local development.

Fixes #21
Fixes #7


.github/workflows/api.yml
```diff
@@ -21,6 +21,8 @@ jobs:
         run: npm run test:coverage
       - name: LINTing
         run: npm run test:lint
+      - name: OpenAPI Schema
+        run: npm run test:schema
```
patricksimonian, line 25: very cool, i'im going to look into this :)

api/src/controllers/namespace.ts
```diff
@@ -0,0 +1,131 @@
...
+export const createNamespace = async (req: Request, res: Response): Promise<void> => {
+  const dm = new DataManager(config);
+  const { NamespaceModel } = dm;
+  const { profileId } = req.params;
+  const { body } = req;
+  const aBody = { ...body, profileId };
+
+  const diff = difference(NamespaceModel.requiredFields, Object.keys(aBody));
+  if (diff.length !== 0) {
+    throw errorWithCode(`Missing required properties: ${diff}`, 400);
```
patricksimonian, line 36: does that output correctly or `properties:  {Object object}`?  if not perhaps `diff.join()`

api/src/controllers/namespace.ts
```diff
@@ -0,0 +1,131 @@
...
+};
+
+export const updateProfileNamespace = async (req: Request, res: Response): Promise<void> => {
+  const dm = new DataManager(config);
+  const { NamespaceModel } = dm;
+  const { profileId, namespaceId } = req.params;
+  const { body } = req;
+
+  // Make sure these are not updated!
+  delete body.id;
```
patricksimonian, line 93: i think performance, al thouguht not critical here, would prefer  spreading new properties into a new object vs delete. 
also delete is a mutation of an existing js, where is typically immutably treating objects is the better practice. 
```
const { id, archived, createdAt, updatedAt, ...body} = req;
const aBody = {...body, profileId};
```

db/sql/0001.sql
```diff
@@ -8,12 +8,29 @@ BEGIN
 END;
 $$ language 'plpgsql';
 
+CREATE TABLE IF NOT EXISTS ref_cluster (
```
patricksimonian, line 11: just for my knowledge is this a form of a migration?

I honestly like this more than using something like knex.js or sequelize

api/src/controllers/namespace.ts
jleach: I've tested it and it seem to be working as expected.

api/src/controllers/namespace.ts
jleach: I've created a bug to go back and clean this up here and another place.

db/sql/0001.sql
jleach: It can read that way. Since I don't expect the schema to exists `CREATE TABLE` would be fine. Its just a habit in case its run multiple times (it won't fail) but it also won't make any changes. Migrations would generally have a `migrations` table in the DB and as the last line in the SQL insert the name of the script into the table to note it was already run.

PR #29, opened 2020-05-15
Use the gp pool to make queries

### LINKED ISSUE #28: Fix database pool usage
The db needs to extract a client from the Pool to run a query then release it back to the pool when done.

jleach:
This changes the way queries are made against Pg. Now clients are taken from the Pool and used to make a query then release back to the pool.

Fixes #28


api/src/db/model/model.ts
```diff
@@ -47,13 +50,28 @@ export abstract class Model {
...
   }
+
+  async runQuery(query: Query): Promise<any[]> {
+    let client;
+
+    try {
+      client = await this.pool.connect();
+      const results = await client.query(query);
+
+      return results.rows.map(r => transformKeysToCamelCase(r));
```
patricksimonian, line 70: not a big deal at all but this could be called `return results.rows.map(transformKeysToCamelCase);` just the same as a fyi

jleach: Good point. Will update.

PR #45, opened 2020-06-23
Requisition form for namespace provisioning

jleach:
This is the initial implementation of the requisition form for namespace provisioning.


web/src/components/UI/button.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
+
+export interface IButtonProps {
+    children?: React.ReactNode,
+    onClick?: (e: any) => void
+};
+
+const styles = {
```
patricksimonian, line 8: my suggestion here is to leverage rebass w/ emotion over a custom button component. Rebass has a button component that you can customize. 

```js
import { Button as RebassBtn} from 'rebass';
import styled from '@emotion/styled';

export const Button = styled(RebassBtn)`
...styles here (keep in mind these are true css properties)
`;


That or you can leverage `emotion-themeing` and create Button variants which will more closely accomplish what you have here.

File: web/src/components/UI/shadowContainer.tsx [hunk 2]
```diff
@@ -0,0 +1,12 @@
+import styled from '@emotion/styled';
+import { Box, Flex } from 'rebass';
+
+
+const boxShadow = `box-shadow: 1px 3px 4px 4px #cbcbcb7d`;
```
patricksimonian, line 5: If this is a common box shadow, than one way of nicely organizing reusable styled components is to place them into a file called `styleTokens.js` and just have them exportable and then import them into files such as this. I use style tokens for common styles like `border-radius`, `border`, `box-shadow` etc.

web/src/components/footer.tsx
```diff
@@ -0,0 +1,86 @@
...
+const StyledFooter = styled.footer`
+  background-color: #036;
+  border-top: 2px solid #fcba19;
+  color: #fff;
+  display: flex;
+  flex-flow: row wrap;
+  text-align: center;
+  position: fixed;
+  bottom: 0px;
+  height: 46px;
```
patricksimonian, line 33: fixing a height can run you into issues later. Its best to often just add padding to make the 'expected' height. For example if text wraps on screen resizes it can look wonky with a fixed height.

web/src/components/form.tsx
```diff
@@ -0,0 +1,292 @@
...
+    };
+
+    // if (keycloak && keycloak.authenticated) {
+    //     console.log('XXXXXXXXXXXX')
+    //     axi.defaults.headers.common['Authorization'] = `Bearer ${keycloak?.token}`;
+    // }
+
+    useEffect(() => {
+        async function wrap() {
+            const result = await axi.get('ministry', {
```
patricksimonian, line 119: a good pattern is to setstate for a loading indicator as well as errors so that you know when they happen. Now if theres like a global loading indicator you can set it using some type of higher order component. This is how global snackbars work

web/src/components/form.tsx
```diff
@@ -0,0 +1,292 @@
...
+            });
+
+            if (result.data) {
+                console.log('Fetched ministry sponsors!!!');
+                setState(result.data);
+            }
+        }
+
+        wrap();
+    }, []);
```
patricksimonian, line 133: Is ministry sponsors relatively static? If it is then calling once is fine, but if this is going too change based on what users do there may need to be some extra logic to fetch them again when the data goes stale

PR #48, opened 2020-07-13
Record details of user who created a project profile

### LINKED ISSUE #39: The registry should link projects to the account that created them
When request is submitted via the API the user will be authenticated against their SSO account. It should be recorded who requested the project set be created.

jleach:
When a project profile is created the KeyCloak ID is recorded so that we determine who made the request.

Fixes #39


api/src/db/model/userprofile.ts
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  async update(userProfileId, data: UserProfile): Promise<UserProfile> {
+    const values: any[] = [];
+    const query = {
+      text: `
+        UPDATE ${this.table}
+          SET
+            last_seen_at = $1
+          WHERE id = $5
```
patricksimonian, line 71: $5? not $2?

jleach: Good catch. Fixed.

PR #51, opened 2020-07-20
Add mechanics to trigger namespace provisioning

### LINKED ISSUE #32: Send message to NATs to trigger namespace provisioning
When a project namespace set is requested the API should send a message to NATS to trigger the provisioning workflow. Draft mock of the messages is:

```
{
  "id": 1,
  "projectName": Health Gateway",
  "namespaces": [
     {
      "displayName": "Health Gateway (tools)",
      "name": "tools-abc123",
      "clusters": [
        "blarb",
        "foo"
      ]
    }
  ]
}
```

Where `id` is the reference to the project record in the registry. It should be added as the label `registryId` to the namespace metadata.

jleach:
This PR adds the necessary mechanics to trigger the namespace provisioning workflow. This is done by creating a JSON payload and submitting it to a NATS subscription.

Fixed #32


api/src/db/model/profile.ts
```diff
@@ -139,4 +144,26 @@ export default class ProfileModel extends Model {
       throw err;
     }
   }
+
+  async isNamespacePrefixUnique(prefix: string): Promise<boolean> {
+    const query = {
+      text: `
+        SELECT COUNT(*) FROM ${this.table}
```
ShellyXueHan, line 151: is the plan to fill in all the existing namespaces to db first?

api/src/db/utils.ts
```diff
@@ -32,3 +33,10 @@ export const transformKeysToCamelCase = data => {
 
   return obj;
 };
+
+export const generateNamespacePrefix = (len: number = 7): string => {
```
ShellyXueHan, line 37: length is 6 digits so far

api/src/libs/fulfillment.ts
```diff
@@ -0,0 +1,104 @@
...
+    const contact = contacts.filter(c => c.roleId === 2).pop();
+
+    if (!contact) {
+      return;
+    }
+
+    const context = {
+      profileId: profile.id,
+      displayName: profile.name,
+      namespaces: [
```
ShellyXueHan, line 51: fake data?

api/src/db/model/profile.ts
jleach: The prefix will be the licence plate we have now. This fn checks to make sure the newly minted one does not already exists in the db; its super unlikely that it would given its 7 random characters but its best to check to make sure.

api/src/db/utils.ts
jleach: Fixed.

api/src/libs/fulfillment.ts
jleach: Good catch. Its not meant to be. Fixing...

jleach: Fixed.

PR #70, opened 2020-09-22
Convert the Web application from a single page app to multipage

yhfreeman:
- adding react ReachRouter lib on the web side of code to support multipage
- modifying the existing view of form submission so it falls into '/' route

#71


web/src/components/AppRouter.tsx
```diff
@@ -0,0 +1,16 @@
+import { Router } from '@reach/router';
```
jleach, line 1: @yhfreeman Any new files need the to have the licensing header at the top; i've forgotten it in a few places but we should fix it going fwd. Check out the header at the top of constants.ts as an example.

web/src/hoc/withLayout.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
```
jleach, line 1: header

jleach (approved):
Just need the header at the top of a few of your new files.

PR #86, opened 2020-10-02
switch to a better approach where dropdown menu visibility is managed by React

yhfreeman:
- removing the workaround which is manipulating the dropdown menu visibility directly from DOM
- implementing changes with the react ref approach where the dropdown menu visibility is now managed by React
- adding TODO note to fix tsx compile error


patricksimonian (approved):
🔥  nice work!

PR #99, opened 2020-10-07
Update NATS message body format and include PO details

### LINKED ISSUE #96: Send the email address of the PO and TC in the nats message
## Story

As a platform administrator (PS) I would like the email for the PO and TC sent as part of the nats message.

## Description

Some information provided in the NATS messages is added to the namespace as metadata. One of the items we're currently including on OCP 3.11 is the email of the PO and TC. To keep this working on OCP4 we need this information as part of the NATS message. This will require some changes to the NATS message body. For example

Current
```json
    technicalContact: { githubId: 'jleach' } }
```

Proposed
```json
   technicalContact: { userId: 'jleach', provider: 'idir', email: 'me@example.com' }
```

## Done When

The NATS message includes the email contact of the TC and PO.

jleach:
- Change nats message body
- Remove magic numbers

Fixes #96

### DIFF
--- a/api/src/constants.ts
+++ b/api/src/constants.ts
@@ -32,7 +32,7 @@ export const projectSetNames = [
   'prod',
 ];
 
-export const ROLE_ID = {
+export const ROLE_IDS = {
   PRODUCT_OWNER: 1,
-  TECHNICAL_STEWART: 2,
+  TECHNICAL_CONTACT: 2,
 };
--- a/api/src/constants.ts
+++ b/api/src/constants.ts
@@ -31,3 +31,8 @@ export const projectSetNames = [
   'test',
   'prod',
 ];
+
+export const ROLE_ID = {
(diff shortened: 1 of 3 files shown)


api/src/constants.ts
```diff
@@ -31,3 +31,8 @@ export const projectSetNames = [
   'test',
   'prod',
 ];
+
+export const ROLE_ID = {
+  PRODUCT_OWNER: 1,
+  TECHNICAL_STEWART: 2,
```
sbathgate, line 37: TECHNICAL_STEWART is used as the constant, however Technical contact is used everywhere else. Not sure if one has been decided as the official title?

sbathgate (approved):
Obviously the "provider": "github" has been hardcoded and will need updated as per your note. Otherwise, it all LGTM.

Issue #76, opened 2020-09-28
Capture component metadata for Platform projects in Project Registry


mitovskaol:
## Story

As an executive I would like to know what frequently used components Platform projects contain in order to be more effective with directing the work on common components for the Platform community and focus on components that are used most commonly first.

## Description

Right below where the user selects the ministry sponsor for the project, lets add the following **optional** check boxes. The users should be prompted to check what services they plan to implement or utilize as part of their project. We can make it a two column so as to not make the form overly long.

Please indicate what services you expect to utilize as part of your project?

 Notification: Email 
 Notification: SMS
 Notification: MS Teams
 Payment processing: Bambora
 Payment processing: PayBC
 File Transfer
 Fire Storage
 Geo Mapping: Web-based
 Geo Mapping:  Location Services
 Scheduling: Calendar
 Scheduling: Apointments
 Identity Management: SiteMinder
 Identity Management:  KeyCloak
 Identity Management: Active Directory
 Other: with an input field

# Done When

The user can fill out the form and the above information is captured by the Web form, sent as part of the POST request, and then processed and recorded by the API.

PR #101, opened 2020-10-08
Capture component metadata for Platform projects in Project Registry

yhfreeman:
- adding optional checkboxes & an input field in order to capture component metadata according to https://github.com/bcgov/platform-services-registry/issues/76
- hooking the values to api call of creating a new profile

**Notes:**
- The current form width doesn't seem to accommodate two columns of checkboxes. We could make adjustments - so this part is open to team discussion.


web/src/constants.ts
```diff
@@ -38,4 +38,21 @@ export const DEFAULT_MINISTRY = 'CITZ';
...
+  { displayName: 'Notification: MS Teams', inputValue: 'notificationMSTeams' },
+  { displayName: 'Payment processing: Bambora', inputValue: 'paymentBambora' },
+  { displayName: 'Payment processing: PayBC', inputValue: 'paymentPayBC' },
+  { displayName: 'File Transfer', inputValue: 'fileTransfer' },
+  { displayName: 'File Storage', inputValue: 'fileStorage' },
+  { displayName: 'Geo Mapping: Web-based', inputValue: 'geoMappingWeb' },
+  { displayName: 'Geo Mapping: Location Services', inputValue: 'geoMappingLocation' },
+  { displayName: 'Scheduling: Calendar', inputValue: 'schedulingCalendar' },
+  { displayName: 'Scheduling: Appointments', inputValue: 'schedulingAppointments' },
+  { displayName: 'Identity Management: SiteMinder', inputValue: 'identityManagementSiteMinder' },
```
sbathgate, line 55: Just a note, in my last commit `identityManagement` was changed to `idm` as per my PR review. So `identityManagementSiteMinder` will be `idmSiteMinder`. This is carries for `idmKeyCloak` and `idmActiveDir`

This can be seen in the [Profile controller](https://github.com/bcgov/platform-services-registry/blob/featuer/issue-76/api/src/controllers/profile.ts#L146-L148)

sbathgate (changes requested):

Issue #49, opened 2020-07-13
The API should have the concept of access control


jleach:
## Story

Ref: See the story in #73.

## Description

The API should implement access control as follows:

1) A regular user can only access their own project profiles.
2) Admin users have read access to all project profiles (and subsequent resources).

## Done When

The API only returns the project profiles associated with the user who's JWT is passed in with a request.

Issue #73, opened 2020-09-22
Create a dashboard where the user can view their existing projects.


jleach:
## Story

As a user, I would like to see all my previously requested projects in a "Dashboard" format so that I can see what I have requested and I have a clear understanding of what requests were received by the registry.

## Description

When a user authenticates they should be taken to a "Dashboard" where they can view a table of all the namespaces they have requested. 

![Screen Shot 2020-09-22 at 2.56.44 PM.png](https://images.zenhubusercontent.com/5c2fa1b7ce4f66381cd8d02d/399b8eae-1b93-4310-9bb1-8de53007b381)

## Done When

A user logs into the dashboard and sees a history of their requests (and only their requests) in table format with each "card" representing a project.

jleach:
@yhfreeman Hey, I'll make some API change for this ticket so that it only returns the authenticated users projects and not every project in the registry (which is what will happen now).

mitovskaol:
Can we please display the PO and TC email info as well.

mitovskaol:
Justin and I reviewed the dashboard (view-only feature) and are very happy with it. Consider the UAT testing done :)

yhfreeman:
The dashboard view feature is merged in PROD-READY (silver) code base on November 05, 2020.

PR #103, opened 2020-10-08
Implement API access control when fetching profiles

sbathgate:
Resolves issue #49. To achieve issue #73 Create a dashboard where the user can view their existing projects, an access control feature is necessary so a regular user can only access their profiles, while an administrator can access all profiles.

This PR implements an access control capability when requesting access to profiles:

- [X] Update fetchAllProjectProfiles() to include access control
- [X] Update fetchProjectProfile() to include access control
- [X] implement findProfilesByUserID() function
- [x] profile unit testing
- [x] update API docs


api/src/controllers/profile.ts
```diff
@@ -67,15 +73,20 @@ export const fetchAllProjectProfiles = async (req: Request, res: Response): Prom
...
+  { params, user }: { params: any, user: AuthenticatedUser }, res: Response
 ): Promise<void> => {
   const { ProfileModel } = dm;
   const { profileId } = params;
 
   try {
     const results = await ProfileModel.findById(Number(profileId));
 
-    res.status(200).json(results);
+    if (user.id === results.userId || user.roles.includes('administrator')) {
```
jleach, line 84: I think I'd put "administrator" into `constants.ts`.

jleach, line 84: I generally prefer early returns to if/else where possible; I think it makes the code more readable and less error prone. For example, one could:
```
if (user.id !== results.userId || !user.roles.includes('administrator')) {
   throw errorWithCode('Unauthorized Access', 401);
}
```

And skip the else.

api/src/controllers/profile.ts
```diff
@@ -67,15 +73,20 @@ export const fetchAllProjectProfiles = async (req: Request, res: Response): Prom
...
+      res.status(200).json(results);
+    } else {
+      throw errorWithCode('Unauthorized Access', 401);
```
jleach, line 87: I don't think this error will make it out. It'll get caught by the catch below and replaced with `throw errorWithCode(message, 500)`; You probably need to check and re-throw.
```
if (err.code) {
  throw err
}
```

jleach (changes requested):
Looking good. I think as part of this ticket you should do all the other endpoints. For example,  `updateProfileNamespace`. Lets get all the CRUD ops done in this one ticket.

api/src/controllers/profile.ts
```diff
@@ -51,11 +51,17 @@ export const uniqueNamespacePrefix = async (): Promise<string | undefined> => {
...
+  { user }: { user: AuthenticatedUser }, res: Response): Promise<void> => {
   const { ProfileModel } = dm;
 
   try {
-    const results = await ProfileModel.findAll();
+    let results;
+    if (user.roles.includes('administrator')) {
+      results = await ProfileModel.findAll();
+    } else {
+      results = await ProfileModel.findProfilesByUserID(user.id);
```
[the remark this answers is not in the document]
sbathgate: Good catch. Adjusted to match camelCase use throughout site. Additionally caught other inconsistencies which I updated accordingly.

jleach (approved):

PR #117, opened 2020-10-15
Fix messaging in provisioning callback

jleach:
Fixes an issue where the namespace ID was being used in place of the profile ID.


sbathgate (approved):
Eventually it might be useful to include specific feedback with these error responses considering they have been separated:
```
const profile = await ProfileModel.findByPrefix(prefix);
if (!profile) {
    throw new Error();
}

const namespaces = await NamespaceModel.findForProfile(Number(profile.id))
if (!namespaces) {
    throw new Error();
}
```

Otherwise LGTM.

PR #126, opened 2020-10-16
Add NSP for API to talk to SSO on the OCP3 cluster

jleach:
Add an external network and config so the API can talk to OCP3.


yhfreeman (approved):
DB service object identity param is updated so that we can pass a label set. It is because we don't always have statefulset pod for DB in other envs such as dev or test.

PR #127, opened 2020-10-16
The project name should not allow any special characters

### LINKED ISSUE #116: The project name should not allow any special characters
## Story

As a system owner, I don't want any special characters in the Project Name

## Description

Only allow letters or numbers in the project name field of the registry form. It has a tendency to break provisioning.

## Done When

The user can not enter punctuation or other special characters (emojis) in the project name field.

yhfreeman:
- adding form validation on project name field so it does not allow any special characters on UI
- adding form validation on other fields

**Form validations:**
- profile name
  - required
  - max 40 chars
  - must be alphanumeric starting with a letter
- profile description:
  - required
  - max 512 chars
  - no special characters that will violate a regular text description
- profile component metadata others:
  - not required
  - max 512 chars
  - no special characters that will violate a regular text description
- contact first name, last name
  - required
  - max 32 chars
  - must be alphabetic
- contact email
  - required
  - max 32 chars
  - must be valid email format
- contact github id
  - required
  - max 32 chars
  - must be valid github id format

 Fixes #116


jleach:
@yhfreeman What was the validation? I think we need (ATM) 40 characters max for the Name and 512 max for the description. @sbathgate we can either ask the Web or API to store the userId in lower case only. Although **people should be able to enter them in mixed case in the Web form**.

sbathgate:
It might be worth converting the userId to lower case within the API in the event calls come from outside the Web interface?

jleach:
Good point. Are you able to create a `bug` ticket for this. See the format we're using in other tickets. Hold off in implementing it; it might be something @j-pye should implement on his end as its only SSO that requires it to be in lower case.

yhfreeman:
@jleach To answer your question "what's the validation", this file defines the constraints https://github.com/bcgov/platform-services-registry/blob/63eaba1b0ff27ffddf9d2e69a9785608c343819d/web/src/utils/useValidator.tsx. I did look up the DB schema constraints, and then set up the front-end validations accordingly.

jleach (approved):

Issue #136, opened 2020-10-20
The API should be able to fetch contacts for a given profile


jleach:
## User Story

As the user, I would like to see the dashboard cards populated with the TC and PO first and last name.

## Done When

sbathgate:
Feature merged and provisioned on registry-web-platform-registry-dev: November 2, 2020.

PR #143, opened 2020-10-26
Add fetch project contacts API capability

sbathgate:
Resolves issue #136.  In order to provide the Product Owner and Technical contact details for the Registry dashboard, the registry API required an additional GET route to return these details.

This PR implements returns the PO and TC details that pertains to the requested profile id:

- [X] Add fetchProfileContacts() to return PO and TC details
- [x] Add profile contacts unit testing
- [X] update API docs


yhfreeman (approved):
All looks good. I understand we have a [separate ticket](https://github.com/bcgov/platform-services-registry/issues/145) for modularizing the authorization part where we do the access control by user_id, so that's good.

PR #156, opened 2020-11-05
Create a dashboard where the user can view their existing projects

### LINKED ISSUE #73: Create a dashboard where the user can view their existing projects.

yhfreeman:
- web side:
  - adding dashboard view feature and related component snapshots
  - updating libraries to remove high vulnerabilities and to support jest test on the web side
- api side:
  - fixing it so it can properly reads the list of roles based on the given jwt payload structure

Asks repo-mountie to ignore the length as there are lengthy web dependencies updates as a result of `npm audit fix`.
Fixes #73. 

/bot-ignore-length


yhfreeman:
Note to self: "Once the new feature moves to prod, please add one last comment to the ticket to say when the feature was implemented and in what cluster, before you close the ticket" from Olena.

sbathgate (approved):
This all looks good to me. Ran everything locally on my system, including my pre-populate bugfix and it works exactly as expected.

api/src/libs/authmware.ts
```diff
@@ -45,6 +45,15 @@ export const isAuthorized = jwtPayload => {
   return true;
 };
 
+export const getJwtPayloadRoles = (jwtPayload: object): string[] | [] => {
+  try {
+    /* tslint:disable-next-line */
+    return jwtPayload['resource_access']['registry-web']['roles'];
```
jleach, line 51: I'd move this client ID to "constants" so it can be changed there as needed.

jleach (approved):

Issue #145, opened 2020-10-29
Add authorization verification function


sbathgate:
## Story
The authorization checks within the API are repetitive, and could be DRY'er. 

## Description
Every API method that requires an authorization check has an individually coded assessment.

## Done When
The authorization functionality is contained within a separate function to improve code re-usability.

sbathgate:
Feature implement in registry-web-platform-registry-dev silver cluster: November 9, 2020.

PR #159, opened 2020-11-06
Refactor profile authorization checks to external function

sbathgate:
As per issue #145 the profile authorization checks were repetitive. This PR extracts the access control validation to an external function to enable future re usability and expansion.


yhfreeman (approved):
Hi I did take a look and it's working well :thumbsup:  Alternatively, we could set it up from the routes perspective by chaining promises, so for further protected routes where we want to ask if a requester is entitled to view / operate around a profile and its related records, we could put the repeated portion in a promise, if not authorized we throw error, if yes we call next() to those exported promises from controller.

PR #194, opened 2020-11-16
Update namespace provisioning email

### LINKED ISSUE #185: Update the email message to indicate who has access to a namespace
## Story

As a PO, TC or Requestor I would like to be informed via the automatic email who has access to a provisioned namespace.

## Description

When a namespace is provisioned a email notification is sent out with some basic details. In this message, add text to indicate that both the PO and TC have been given admin access to the newly minted namespaces.

## Done When

The body of the email indicates all people who have admin access to the new namespaces.

sbathgate:
Users reported concerns regarding a lack of project specific information within the provisioning email. This PR ensures provisioning emails contain the project name, and PO and TC contact details.

This PR includes: 
- [X] Add project name in the email subject
- [X] Add PO and TC details submitted with the project
- [x] Refactor and clean up code
- [x] Update tests to reflect changes

Fixes #185

### DIFF
--- a/api/src/libs/messaging.ts
+++ b/api/src/libs/messaging.ts
@@ -61,20 +61,37 @@ export const profileDetails = async (profileId: number): Promise<string> => {
   }
 }
 
-export const inputVariables = async (buff: string, to: string[], profileName: string, names: string[]): Promise<any> => {
-  buff = buff.replace('${POName}', names[0]);
-  buff = buff.replace('${TCName}', names[1]);
-  buff = buff.replace('${POEmail}', to[0]);
-  buff = buff.replace('${TCEmail}', to[1]);
-  buff = buff.replace('${NamespaceName}', profileName);
-  return buff;
+export const updateEmailContent = async (buff: string, to: string[], profileName: string, contactNames: string[]): Promise<string> => {
+  try {
+    let emailContent: string;
+
+    const mapObj = {
+      POName: contactNames[0],
+      TCName: contactNames[1],
+      POEmail: to[0],
+      TCEmail: to[1],
+      projectName: profileName,
+    };
+
+    const re = new RegExp(Object.keys(mapObj).join('|'),'gi');
+    emailContent = buff.replace(re, matched => {
+      return mapObj[matched];
+    });
+    return emailContent;
+  } catch (err) {
+    const message = `Unable to update email content`;
+    logger.error(`${message}, err = ${err.message}`);
+
+    return '';
+  }
+
 }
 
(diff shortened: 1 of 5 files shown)


api/src/libs/messaging.ts
```diff
@@ -46,10 +45,55 @@ export const contactsForProfile = async (profileId: number): Promise<string[]> =
...
 }
 
+export const profileDetails = async (profileId: number): Promise<string> => {
+
+  try {
+    const dm = new DataManager(shared.pgPool);
+    const { ProfileModel } = dm;
+    const profiles = await ProfileModel.findById(profileId);
+    const profileName = profiles.name;
+    return profileName;
```
jleach, line 55: Could probably do with one space in here just before the return. I think the only advantage would be to break the code up into "stanzas" for lack of a better work to improve readability.

api/src/libs/messaging.ts
```diff
@@ -46,10 +45,55 @@ export const contactsForProfile = async (profileId: number): Promise<string[]> =
...
+    const mapObj = {
+      POName: contactNames[0],
+      TCName: contactNames[1],
+      POEmail: to[0],
+      TCEmail: to[1],
+      projectName: profileName,
+    };
+
+    const re = new RegExp(Object.keys(mapObj).join('|'),'gi');
+    emailContent = buff.replace(re, matched => {
```
jleach, line 77: I wonder if this would work:

```
emailContent = buff.replace((re, matched) => mapObj[matched]);
```

api/src/libs/messaging.ts
```diff
@@ -46,10 +45,55 @@ export const contactsForProfile = async (profileId: number): Promise<string[]> =
...
+      return mapObj[matched];
+    });
+    return emailContent;
+  } catch (err) {
+    const message = `Unable to update email content`;
+    logger.error(`${message}, err = ${err.message}`);
+
+    return '';
```
jleach, line 85: It **may** be better to do a return type of `Promise<string | undefined>` and then just return rather than returning an empty string. That way you can check if the return value is `undefined` rather than an empty string.

jleach (approved):

api/src/libs/messaging.ts
sbathgate: This would need a minor adjustment as .replace() flags as only one argument if `(re, matched)` is wrapped in (). However, removing that appears to return the keys vs values. Not entirely sure why to be honest though.

sbathgate: I think it was an old bug. Rebuilt the api and have had no further issues.

PR #196, opened 2020-11-17
Project Registry Description field 'Invalid Format' when return key is used

### LINKED ISSUE #166: Project Registry Description field 'Invalid Format' when return key is used
-Description field shows 'Invalid Format' message when return key is used. 
-'Invalid Format' message is not very specific, leaves the user wondering what they did wrong.

yhfreeman:
- providing a fix so the return key is accepted by profile form submission - description field
- updating the dashboard profile card so description can read return key and render new line
- updating snapshots
- renaming useValidator to getValidator to follow the naming conventions where useXXX is only for react state / hooks related stuff

Fixes #166

### DIFF
--- a/web/src/components/UI/newLineText.tsx
+++ b/web/src/components/UI/newLineText.tsx
@@ -1,0 +1,29 @@
+//
+// Copyright © 2020 Province of British Columbia
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


web/src/components/UI/newLineText.tsx
```diff
@@ -0,0 +1,29 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import React from 'react';
+
+const NewlineText = (props: any) => {
```
jleach, line 19: Could add a typescript return type here `string | null`

jleach (approved):

PR #201, opened 2020-11-24
Implement Registry api quota edit feature

### LINKED ISSUE #171: Refactor NATS messaging to accommodate new format
## Story

As a developer, the NATS message should be updated to support the new message format (see dependency).

## Description

The NATS message will be refactored for both requesting new namespaces (existing functionality) and to allow for extension for edit. For this story, we just need to make the **existing** functionality work with the new format - not the new edit functionality.

## Done When

A new namespace request can be made with and the new message format is sent.

### LINKED ISSUE #172: Implement new NATS message for edit functionality
## Story

As a developer, I would like the new "edit" message to be sent as needed

## Description

The NATS message redesigned as part of this ticket (see dependencies) will require either two separate message or one message with a `type` to indicate if it's for an "edit" or "new project set". This story is to implement the message sent when "edit".

## Done When

A properly formated message is sent to NATS when a project "edit" is being requested.

### LINKED ISSUE #174: Implement request tracking so that edits are clear and concise
## Description

Ref. See meeting notes in associated EPIC.

When an edit is requested, similar type of data should be packaged into a single request. For example, if the PO is changed, then the TC is changed, it would be good if both can be packaged so that only one request is sent to the robots for implementation. In addition to this, some requests must be processed sequentially and **must bock** subsequent requests. For example, a user can only progress through CPU request from S to M then to L; they cannot go directly from S to L. If a request to go from S to M is pending, then subsequent quota edits must be blocked. However, it would be possible for the user to edit project metadata (that only exists in the database) or project details like description or TC.

For simplicity, v1 of this feature can treat any edit as the same and block all subsequent edits until said request is made. If implemented this way crate additional tickets to improve later.

The smart robots processing request must apply them in order (FIFO).

## Done When

A user can request edits to quota, and is blocked from making additional requests until the first request is made;
A user can request edits to project details, these requests do not block additional requests to project details.

### LINKED ISSUE #177: Refactor the database so that it can track requests and quotas
## Story

As a developer, I would like the database to be able to track namespace quotas and requests.

## Description

As part of the EPIC this story belongs to, we need to be able to track the requests that are made (new project set, edit project data, edit quota, and others in the future) as well as the quota for CPU / Mem / Storage for each namespace. Refactor the database so that it is ready for feature implementation. Once refactored it should be able to track the new information as well as work with the current requirements (new project).

## Done When

The database can track requests made along with the status of said request, as well as the quotas for each namespace for all three components.

yhfreeman:
- [x] **Refactor the database so that it can track requests and quotas**
- adding quota size "small", "medium", "large" enums in constants
- schema changes including
  - adding "quota_cpu", "quota_memory", "quota_storage" attributes to "cluster_namespace" table
  - creating "request" table
- preparing a script to run against existing cluster_namespace entries

Fixes #177 

-----------------------------
- [x] **Refactor NATS messaging to accommodate new format**
- modifying namespace model
  - modifying createProjectSet method
    - so that new cluster_namespace entries will be created with default "small" size for cpu, memory and storage quota
  - modifying findForProfile method
    - so that fetched cluster_namespace entries via api will include quota info
- modifying fulfillment nats message context to adapt new format for CREATE namespace

Fixes #171 

-----------------------------
- [x] **Implement request tracking so that edits are clear and concise**
- modifying namespace model
  - adding updateQuota method
- adding request model
  - adding the basic create, update and delete methods
- adding api endpoints / modifying controller so that
  - web client can query for existing edit_quota options with profileId params
GET `http://localhost:8100/api/v1/profile/{{profileId}}/quota-edit`
  - web client can submit edit_quota request with profileId params
POST `http://localhost:8100/api/v1/profile/{{profileId}}/quota-edit`
- nat service can keep using the same callback passing the prefix when quota updates are complete
PUT `http://localhost:8100/api/v1/provision/namespace`

Fixes #174 

-----------------------------
- [x] **Implement new NATS message for edit functionality**

Fixes #172   

-----------------------------
- [x] updating api dependencies

- [x] updating snapshots, unit tests and api doc

due to big feature update, updated dependencies, snapshot and api docs:
/bot-ignore-length


yhfreeman:
This draft PR will be broken down to each individual pieces for the efficiency of code review.

sbathgate (approved):

api/src/router/routes/provision.ts
```diff
@@ -20,12 +20,12 @@
...
 
 const router = express.Router();
 
 // Provisioning
 router.post('/:profileId/namespace', asyncMiddleware(provisionProfileNamespaces));
-router.put('/namespace', asyncMiddleware(updateProvisionedNamespaces));
+router.put('/namespace', asyncMiddleware(provisionCallbackHandler));
 
-export default router;
+export default router;
```
sbathgate, line 31: Nit: No newline
```suggestion
export default router;

```

db/seed/data_migration_0005.sql
```diff
@@ -0,0 +1,7 @@
+
+BEGIN TRANSACTION;
+
+UPDATE cluster_namespace
+  SET quota_cpu = 'small', quota_memory = 'small', quota_storage = 'small';
+
+END TRANSACTION;
```
sbathgate, line 7: Nit: As above, no newline
```suggestion
END TRANSACTION;

```

sbathgate (approved):

api/src/router/routes/provision.ts
yhfreeman: The linter we use seems to require a new line for each file - otherwise it would fail the npm run test:lint.

PR #204, opened 2020-11-24
Bugfix: ensure TCEmail is not undefined

### LINKED ISSUE #203: Bug: TC email shows as undefined
**Describe the issue**
It was observed that if a PO and TC shared the same email address, the namespace provisioning email would display the TC Email as `undefined`.

**Definition of done**
This is complete when the TC email shows the appropriate email address.

sbathgate:
It was discovered if the PO and TC had the same email address, the provisioning email would show the TC Email as `undefined`. This PR checks if the PO and TC have the same address and provisioning the namespace email accordingly.

Fixes #203

### DIFF
--- a/api/src/libs/messaging.ts
+++ b/api/src/libs/messaging.ts
@@ -70,7 +70,7 @@ export const updateEmailContent = async (buff: string, to: string[], profileName
       POName: contactNames[0],
       TCName: contactNames[1],
       POEmail: to[0],
-      TCEmail: to[1],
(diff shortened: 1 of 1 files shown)


yhfreeman (approved):
Alternatively, we could prep "contacts" in the way that emails are strictly retrieved from that as well, instead of replying on "to", the list of recipient emails, with the reason being if we ever want to change who receive the email, the logics around PO/TC info as per profile wouldn't be affected. Thanks.

PR #205, opened 2020-11-25
Refactor the database so that it can track requests and quotas

### LINKED ISSUE #177: Refactor the database so that it can track requests and quotas

yhfreeman:
- adding schema changes including:

  - adding "quota_cpu", "quota_memory", "quota_storage" attributes to "cluster_namespace" table

  - creating "request" table

  - preparing a script to run against existing cluster_namespace entries

- adding quota size "small", "medium", "large" enums in constants

Fixes #177

### DIFF
--- a/api/src/constants.ts
+++ b/api/src/constants.ts
@@ -28,6 +28,14 @@ export const projectSetNames = [
   'prod',
 ];
 
+export const quotaSizeNames = [
+  'small',
+  'medium',
+  'large'
+];
+
+export const DEFAULT_QUOTA_SIZE_NAME = quotaSizeNames[0];
+
 export const ROLE_IDS = {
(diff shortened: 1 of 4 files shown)


yhfreeman:
Closed due to duplicated commits.

PR #234, opened 2020-12-05
Fix the web csv export double quote issue

yhfreeman:
- fixing the web csv export component to resolve issues caused by those profile descriptions that contain double quotes


yhfreeman:
Thanks for catching this bug @juhewitt.
**Bugfix updated on the registry-web-platform-registry-prod silver cluster: December 04, 2020

PR #235, opened 2020-12-07
Current quote edit feature will work well with future profile edits

### LINKED ISSUE #222: Current quote edit feature will work well with future profile edits.
## Story

As a developer, I want to make sure current quota edit feature will work well with other edit features, including profile description and contacts under a profile so there's no conflicts.

## Description

The Registry api code should have necessary and some reusable components that support all edit features in a way that all other edit requests should be blocked when there's an ongoing request for a given profile.

## Done When

There is enough groundwork so we are confident quota edit feature will work well with other edit features.

yhfreeman:
- ts linting
- refactor work around quota edit
- refactor work so there are reusable components that aren't just specific to quota edit but all edits
- refactor work so clusterName is read as optional request param as bot is currently passing to provision endpoint

Fixes #222 

/bot-ignore-length

### DIFF
--- a/api/__tests__/fulfillment.spec.ts
+++ b/api/__tests__/fulfillment.spec.ts
@@ -17,7 +17,8 @@
 import fs from 'fs';
 import path from 'path';
 import { Pool } from 'pg';
-import { contextForProvisioning, fulfillNamespaceEdit, fulfillNamespaceProvisioning, RequestEditType } from '../src/libs/fulfillment';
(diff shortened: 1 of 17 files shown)


sbathgate (approved):
This looks great, will definitely make integrating descriptions and contacts much easier. 

More of a nit pick, but I would have maybe separated the tslint and api updates into individual PR's that way mundane tslint changes don't get held up.

PR #240, opened 2020-12-10
Bugfix for profile description containing unescaped quote in it

### LINKED ISSUE #213: The description can not have unescaped quote in it
## Description

The Web and API/DB can store messages as users enter them, however, when sending the JSON message via NATS the API should:

1. Escape double quotes with `"` -> `\"`.

## Done When

If the description or project name contain single quotes in the Web/DB they are escaped before sending the message to NATS.

yhfreeman:
- on the api side, stringify nats message before sending
- on the web side, allow double quote in profile description

Some notes:
It turns out that the double quotes have been properly escaped when they are collected from the web and stored in DB.
It's just when we send structured data through nats, without JSON.stringify, they won't be escaped properly.
Since the ansible side has already got the step to parse the nats msg, the solution is to JSON.stringify it first.

Fixes #213

### DIFF
--- a/api/src/libs/fulfillment.ts
+++ b/api/src/libs/fulfillment.ts
@@ -92,13 +92,16 @@ const sendNatsMessage = async (profileId: number, natsObject: NatsObject): Promi
     const nc = shared.nats;
     const { natsSubject, natsContext } = natsObject;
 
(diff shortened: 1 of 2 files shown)


sbathgate (approved):
LGTM

PR #253, opened 2020-12-18
Bugfix for unescaped quote issue for endpoint getProfileBotJsonUnderPending

### LINKED ISSUE #213: The description can not have unescaped quote in it

yhfreeman:
This is to resolve an issue that still persists after the bugfix PR #252.

Bugfix PR #252 didn't solve the last case from the list below:
> Actions from Web UI (NATS Used)
> - Create : Quotes are escaped as expected with \\\" in the JSON/NATS message
> - Edit: Quotes are escaped as expected with \\\" in the JSON/NATS message
> 
> Actions from Sync Workflow (API Used) - provisioned-profile-bot-json
> - Create: No Result as Expected (Not Approved/Provisioned Yet)
> - Sync (Provisioned Create or Edit): Quotes are escaped as expected \\\" in the JSON/NATS message
> - Sync (Pending Edit PR Open): Create (Provisioned) JSON/NATS message returned with expected \\\"
> 
> Actions from Sync Pending Workflow (API Used) - under-pending-profile-bot-json
> - Create: Quotes are escaped as expected \\\" in the JSON/NATS message
> - **Edit: Quotes are not escaped properly \\\\\" in the JSON/NATS message**

Therefore, this PR is to accommodate the last case.

Fixes #213

### DIFF
--- a/api/src/controllers/sync.ts
+++ b/api/src/controllers/sync.ts
@@ -96,7 +96,6 @@ export const getProvisionedProfileBotJson = async (
     const stringifiedMsg = JSON.stringify(replaceForDescription(context));
 
     logger.info(`\n\nstringified JSON msg for ${profileId}`, stringifiedMsg, '\n\n');
-    // serve bot the stringified nats/json message to avoid not escaped double quotes issue
     res.status(200).send(stringifiedMsg);
   } catch (err) {
     const message = `Unable get provisioned profile bot json for profile ID ${profileId}`;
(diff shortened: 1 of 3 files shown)


yhfreeman:
Thanks @j-pye for catching the issue and narrowing it down to the cause. Much appreciated.

yhfreeman:
Given the nature of the inconsistent double quote issue, there are concerns over maintaining this proposed bugfix solution going forward. It's decided to adapt a different approach:

Instead of `"` with `\"` or` \\\"`, replace `"` with ` ` .

PR #259, opened 2020-12-24
The status message should let teams know what is pending

### LINKED ISSUE #224: The status message should let teams know what is pending
## Story

As a user, I would like to know what is pending in the Web UI.

## Description

Right now the Web UI just says "pending" when an edit request is in-flight. We would like to to say "Update Pending"

## Done When

The user sees "Update Pending" in the web UI.

yhfreeman:
- update the status message to let teams know what's pending

Fixes #224


sbathgate (approved):
LGTM. Tested locally, and works great!

PR #261, opened 2021-01-04
Manually resolve dependabot alert issue on node-notifier

yhfreeman:
- fix the dependabot alert node-notifier >=8.0.1

To ensure that functionalities and behaviours are working as expected after the libraries update, the code was tested locally by running unit tests for registry api portion using jest, as that's where the node-notifier and its parent library are being used.

Due to dependency and sub-dependency updates:
/bot-ignore-length

### DIFF
--- a/api/package-lock.json
+++ b/api/package-lock.json
@@ -52,42 +52,41 @@
       }
     },
     "@babel/core": {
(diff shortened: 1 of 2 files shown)


sbathgate (approved):
LGTM

PR #272, opened 2021-01-12
Resolve web portal unauth 401 issue after token gets refreshed

### LINKED ISSUE #266: Resolve web portal unauth 401 issue after token gets refreshed
**Describe the issue**
For those authenticated users that idle around the website >= 5mins long, any requests made after the idle will be marked as unauthorized, which is the issue we have. Instead of suggesting users to log out and log in, the bugfix should be to fix the behaviours of the react app so it uses / tracks the refreshed token.

**Definition of done**
When the unauth 401 issue after token gets refreshed is resolved.

yhfreeman:
bugfix for all api calls, state management so as to use the refreshed token

Fixes #266


sbathgate (approved):
With the `updateProduct` adjusted to `put` and the logger now commented out. All LGTM!

PR #285, opened 2021-01-19
Clean up frontend state mgmt including ProfileEdit component

### LINKED ISSUE #276: Clean up frontend state mgmt including ProfileEdit component
**Describe the issue**
As a dev, I'd like to have some cleanup work done around frontend state mgmt in order to make sure the current glitch is resolved and the code is easy to read / maintain.

**Additional context**
ProfileEdit component

- Resolve glitch: when users submit a profile request successfully and are directed back to the dashboard view, the fetched / rendered profile data do not always reflect the changes unless users reload the same page again.
- Clean up states

Other components
- Clean up states

**Definition of done**
The above cleanup work is done.
The above glitch is resolves where users can see reflected changes immediately without having to reload the page.

yhfreeman:
- [x] cleanup around auth state-related props on Layout, AppRoute, PrivateRoute
- [x] add context and hook so backdrop (when data are being processed in the background) state can be reached across components
- [x] cleanup around ProfileEdit

Fixes #276 

Due to refactor volume:
/bot-ignore-length

### DIFF
--- a/web/src/components/profileEdit/ProjectCardEdit.tsx
+++ b/web/src/components/profileEdit/ProjectCardEdit.tsx
@@ -13,6 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
+
 import { Input, Label, Textarea } from '@rebass/forms';
 import React, { useState } from 'react';
 import { Field, Form } from 'react-final-form';
@@ -26,13 +27,16 @@ import { promptErrToastWithText, promptSuccessToastWithText } from '../../utils/
 import { transformForm } from '../../utils/transformDataHelper';
 import { StyledFormButton, StyledFormDisabledButton } from '../common/UI/Button';
 import FormTitle from '../common/UI/FormTitle';
+import { ProjectDetails } from './ProjectCard';
+
+const validator = getValidator();
 
 interface IProjectCardEditProps {
-    profileDetails: any;
+    projectDetails: ProjectDetails;
     ministry: Array<MinistryItem>;
+    handleSubmitRefresh: any;
     isProvisioned?: boolean;
-    pendingEditRequest: boolean;
-    setPendingEditRequest: any;
+    hasPendingEdit?: boolean;
 }
 
 interface MinistryItem {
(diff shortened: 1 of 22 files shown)


yhfreeman:
Deleted repo-mountie length warning.

web/src/utils/getProfileStatus.tsx
```diff
@@ -14,13 +14,13 @@
 // limitations under the License.
 //
 
-
 export default function getProfileStatus() {
-    const getPendingEditRequest = async (api: any, profileId: string): Promise<any> => {
+    const hasPendingEditRequest = async (api: any, profileId: string): Promise<any> => {
```
sbathgate, line 18: Good catch, better choice of verbiage

web/src/components/profileEdit/ProjectCardEdit.tsx
```diff
@@ -132,90 +194,32 @@ const ProjectCardEdit: React.FC<IProjectCardEditProps> = (props) => {
...
+                    ) : (
+                            <>
+                                {/* @ts-ignore */}
+                                <StyledFormDisabledButton style={{ display: 'block' }}>Request Update</StyledFormDisabledButton>
+                                <Label as="span" variant="errorLabel" >Not available due to a {isProvisioned ? 'Update' : 'Provision'} Request </Label>
+                            </>
+                        )
+                    }
+                </form>
+            )}
```
sbathgate, line 221: Note for later: This block is identical between ProjectCardEdit and ContactCardEdit, only additions for QuotaCardEdit is use of term 'Quota' instead of 'Update' and it has `onClick={handleSubmit}`. This could likely be refactored to reduce the duplication.

web/src/utils/transformDataHelper.tsx
```diff
@@ -166,10 +166,10 @@ export function transformJsonToCsv(objArray: any) {
   return str;
 };
 
-export function getProfileMinistry(ministrySet: any[], profileDetails: any): object {
+export function getProfileMinistry(ministrySet: any[], projectDetails: any): object {
   let ministryDetails: any = {};
   ministrySet.forEach((ministry: any) => {
-    if (ministry.id === profileDetails.busOrgId) {
+    if (ministry.id === projectDetails.busOrgId) {
```
sbathgate, line 172: Creating consistency with 'project' vs 'profile' was a great update! 👍

sbathgate (approved):
Other than the double props call as we have discussed this looks great.

web/src/components/profileEdit/ProjectCardEdit.tsx
yhfreeman: sounds good - can refactor

yhfreeman:
Hi @sbathgate I just pushed [this commit](https://github.com/bcgov/platform-services-registry/pull/285/commits/fb1ab2d65a5ab2a8a1aaa80f8927a97f3d5d6cab) because I was previously mistaken that the logic (of checking whether an edit request should go through bot or not) was on the api side. It's fine if that piece of logic is currently handled by the web ui, but perhaps we could move it to the api in the future?

PR #287, opened 2021-01-25
Both PO and TC user names should be passed to NATS in lower case

### LINKED ISSUE #227: Both PO and TC user names should be passed to NATS in lower case
## Story

As a bot developer, I need the username for both the PO and TC to be in lower case.

## Description

The API should convert the PO and TC usernames to lower case in the NATS message.

## Done When

All messages sent to NATS have the usernames in lower case.

yhfreeman:
- ensure data written to db for username and email are lower case
- prep a sql to ensure all the past data written in db for username and email are lower case

the above implementations will ensure all PO, TC usernames are lower case in nats message

Fixes #227

### DIFF
--- a/api/src/db/model/contact.ts
+++ b/api/src/db/model/contact.ts
@@ -51,7 +51,7 @@ export default class ContactModel extends Model {
       values: [
         data.firstName,
         data.lastName,
-        data.email,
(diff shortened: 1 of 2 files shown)


sbathgate (approved):
LGTM

PR #292, opened 2021-01-28
Fix frontend routes so logout is working without throwing errors

### LINKED ISSUE #290: Fix frontend routes so logout is working without throwing errors
**Describe the issue**
As a dev, I'd like to have frontend routes working as expected so the logout functionality is working without throwing errors.

**Additional context**
If users go to the service while logged out and click on the main logo, it tries to get users to goto the dashboard which throws an error.

If users login to the registry and then logout, it throws another error as it again tries to route users to the dashboard which is obviously unauthorized now.

**Definition of done**
When the above is fixed.

yhfreeman:
- bugfix so that the logout functionality is working without throwing errors

Fixes #290

### DIFF
--- a/web/src/AppRouter.tsx
+++ b/web/src/AppRouter.tsx
@@ -40,11 +40,17 @@ const AppRouter: React.FC = () => {
   return (
     <Router history={browserHistory}>
       <Switch>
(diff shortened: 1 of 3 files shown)


sbathgate (approved):
LGTM

PR #294, opened 2021-02-03
Some groundwork changes prior to implementing quota refactor

yhfreeman:
- Schema changes.
  - Add new table `ref_quota`.
  - Add `primary_cluster_name` to existing table `profile`.
    - This is because default cluster might change in the future, so we need a place to store what was being used when a new CREATE request is made.
    - So far, we have logics spread among the Registry api code that we consider a profile is provisioned should all 4 namespaces on the default cluster are provisioned. With the primary_cluster_name added, we can now abstract these logics and give it a meaning through primary namespace set.

- Add suffix `_size` to `quota_cpu`, `quota_memory`, and `quota_storage` for clarity.
- Remove temp_flyway_migration folder as it has been complete.

/bot-ignore-length


api/src/db/model/namespace.ts
```diff
@@ -77,11 +79,64 @@ export default class NamespaceModel extends Model {
...
+      return results.pop();
+    } catch (err) {
+      const message = `Unable to archive namespace ${namespaceId}`;
+      logger.error(`${message}, err = ${err.message}`);
+
+      throw err;
+    }
+  }
+
   async createProjectSet(profileId: number, clusterId: number, prefix: string,): Promise<ProjectNamespace[]> {
```
sbathgate, line 135:
```suggestion
  async createProjectSet(profileId: number, prefix: string, clusterId?: number): Promise<ProjectNamespace[]> {
```

api/src/controllers/provision.ts
```diff
@@ -43,16 +42,16 @@ export const provisionProfileNamespaces = async (
...
-      if (!clusterId || !profile) {
-        errorWithCode(500, 'Unable to fetch info for provisioning');
+      if (!profile || !cluster) {
+        const errmsg = 'Unable to fetch info for provisioning';
+        throw new Error(errmsg);
       }
 
-      await NamespaceModel.createProjectSet(profileId, clusterId, profile.namespacePrefix);
+      // @ts-expect-error
+      await NamespaceModel.createProjectSet(profileId, cluster.id, profile.namespacePrefix);
```
sbathgate, line 54:
```suggestion
      await NamespaceModel.createProjectSet(profileId, profile.namespacePrefix, cluster.id);
```

sbathgate (approved):
Overall everything looks good. I have included a thought for consideration RE: `// @ts-expect-error`. Given `cluster.id` is typed to the ClusterModel it expects `number | undefined` however in createProjectSet it obviously defines the type as `number` I wasn't sure if implementing the optional tag? would be preferential to ignoring it. Alternatively, we could define the type in createProjectSet as `any`.

PR #297, opened 2021-02-05
Authorization refactor

sbathgate:
This PR refactors how the registry conducts its Authorization processes. The previous system relied exclusively upon the signed in users KeyCloak id, this refactor enables a users email to be a means of authorization also.

### DIFF
--- a/api/src/controllers/profile.ts
+++ b/api/src/controllers/profile.ts
@@ -25,9 +25,10 @@ import DataManager from '../db';
 import { Request, RequestEditType } from '../db/model/request';
 import { generateNamespacePrefix } from '../db/utils';
 import { AuthenticatedUser } from '../libs/authmware';
+import { getAuthorization } from '../libs/authorization';
 import { fulfillNamespaceEdit } from '../libs/fulfillment';
 import shared from '../libs/shared';
-import { isNotAuthorized, validateObjProps } from '../libs/utils';
+import { validateObjProps } from '../libs/utils';
 
 const dm = new DataManager(shared.pgPool);
 const whichService = 'profile editing';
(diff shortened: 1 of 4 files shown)


api/src/controllers/profile.ts
```diff
@@ -273,21 +272,19 @@ export const addContactToProfile = async (
...
-  const { ContactModel, ProfileModel } = dm;
+  const { ContactModel } = dm;
   const { profileId } = params;
 
   try {
-    const record = await ProfileModel.findById(Number(profileId));
+    const projectContacts = await ContactModel.findForProject(Number(profileId));
 
-    const notAuthorized = isNotAuthorized(record, user);
+    const isAuthorized = getAuthorization(profileId, user, undefined, projectContacts);
```
yhfreeman, line 281: Just nitpick on readability. Currently getAuthorization either throw an Error or return true. It's just a bit odd to "throw isAuthorized" when "not authorized"?

api/src/controllers/profile.ts
```diff
@@ -83,15 +84,15 @@ export const fetchProjectProfile = async (
...
+    const projectDetails = await ProfileModel.findById(Number(profileId));
 
-    const notAuthorized = isNotAuthorized(record, user);
+    const isAuthorized = getAuthorization(profileId, user, projectDetails, undefined);
 
-    if (notAuthorized) {
-      throw notAuthorized;
+    if (!(isAuthorized)) {
+      throw isAuthorized;
     }
```
yhfreeman, line 93: Just nitpick on readability. Currently getAuthorization either throw an Error or return true. It's just a bit odd to "throw isAuthorized" when "not authorized"?

yhfreeman (approved):
The feature is functioning as expected - thanks for your work. Perhaps libs/authorization could somehow be be combined with libs/authware. Also, would it be worth cleaning up some linter issues according to sonarcloud?

api/src/controllers/profile.ts
sbathgate: Absolutely agree, I have adjusted to isNotAuthorized to reflect the logic more appropriately.

PR #300, opened 2021-02-06
Quota refactor - web ui

yhfreeman:
- adjustment on the web ui side to accommodate quote refactor
- modify around quote-edit submit button so it's greyed out until users make a selection


sbathgate (commented):
LGTM, more readable and clear than previous too! :)

PR #330, opened 2021-02-27
Migrate application from NSP to KNP

jleach:
Migrate the application from NSP to KNP.

### DIFF
--- a/openshift/templates/nsp.yaml
+++ b/openshift/templates/nsp.yaml
@@ -1,9 +1,9 @@
 apiVersion: template.openshift.io/v1
 kind: Template
 labels:
-  template: mobile-signing-service
+  template: platform-registry
 metadata:
-  name: mobile-signing-service
+  name: platform-registry-network-policies
 objects:
   - apiVersion: security.devops.gov.bc.ca/v1alpha1
     kind: NetworkSecurityPolicy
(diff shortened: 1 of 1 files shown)


jleach:
@sbathgate @yhfreeman I've applied this to `dev`. One tested and approved I'll promote it through to test / prod if you're both ok with that.

yhfreeman:
Hi @jleach on the new knp deny-all basis, we still need to add ingress rule `allow-flyway-to-patroni`.
Tested this on dev the other day - this should work:
```
  - kind: NetworkPolicy
    apiVersion: networking.k8s.io/v1
    metadata:
      name: allow-flyway-to-patroni
    spec:
      # Allow flyway pre-hook pod to run
      # db migration check on patroni
      # each time api deployment rolls out
      podSelector:
        matchLabels:
          statefulset: registry-patroni
      ingress:
        - from:
            - podSelector:
                matchLabels:
                  openshift.io/deployer-pod.type: hook-pre
          ports:
            - protocol: TCP
              port: 5432
```

Also, registry test instance is connected to the nats service in `platform-provisioner-dev` namespace.
So before promoting to test, we might want to add an ingress rule there so api can reach out.

jleach:
@yhfreeman Like the baby-yoga icon. Yes, @j-pye will need to add ingress policy eventually for NATS but we won't need to change anything. There isn't support for egress rules on our side until OCP 4.8. I though fly-away ran inside the API pod for some reason so would be covered by the policy.  Or does it run as its own pod?

yhfreeman:
@jleach Thanks! RE flyway, it's both. You're correct, the one that runs inside the API pod would be covered, but the other one that runs on its own pod (and does the actual db version check and apply new schema if there's any) would need additional ingress rule.

yhfreeman (approved):

PR #370, opened 2021-04-07
Multi-cluster support feature - Registry API changes

yhfreeman:
Due to the unit test snapshot updates
/bot-ignore-length


sbathgate (commented):
Overall, this LGTM, I have tied in a comment on the Multi-cluster support feature - Registry WEB changes that might result in some API adjustments as a result so I will follow up with an approval after we have discussed those.

sbathgate (approved):

PR #371, opened 2021-04-07
Multi-cluster support feature - Registry WEB changes

yhfreeman:
Due to the unit test snapshot updates
/bot-ignore-length


web/src/components/profileEdit/ProjectCard.tsx
```diff
@@ -72,6 +79,17 @@ const ProjectCard: React.FC<IProjectCardProps> = (props) => {
           {ministryName}
         </Text>
       </Box>
+      <Box width={1 / 2} px={2} mt={3}>
+        <Text as="h3">Cluster</Text>
+        <Text as="p" color={theme.colors.grey} fontSize={[1, 2, 2]} mt={1}>
+          This is the cluster name;
```
sbathgate, line 85:
```suggestion
          This is the cluster this namespace is provisioned in;
```

web/src/utils/transformDataHelper.tsx
```diff
@@ -73,14 +73,19 @@ export function sortProfileByDatetime(profileData: any): any[] | [] {
...
   try {
-    namespaceSet.forEach((namespace: any) => {
-      // TODO: update this when we add other clusters besides default cluster
-      if (!namespace.clusters[0].provisioned) {
-        throw Error(`${namespace.name} is not provisioned yet`);
+    const { primaryClusterName } = profile;
+
+    namespaces.forEach((namespace: any) => {
+      const clusterNamespace = namespace.clusters.filter(
+        (c: any) => c.name === primaryClusterName,
```
sbathgate, line 82:
```suggestion
        (cluster: any) => cluster.name === primaryClusterName,
```
I recognize contextually it is obvious that c means cluster, however I wasn't sure if more explicit terminology would be beneficial.

web/src/views/ProfileCreate.tsx
```diff
@@ -71,8 +72,10 @@ const ProfileCreate: React.FC = () => {
 
   useEffect(() => {
     async function wrap() {
-      const response = await api.getMinistry();
-      setMinistry(response.data);
+      const ministryResponse = await api.getMinistry();
+      const clusterResponse = await api.getCluster();
```
sbathgate, line 76: This is returning ALL clusters (ARO for instance isn't being used any more, and it technically allows for someone to provision on Gold OR Gold DR which won't be support as I understand it.

Additionally, there is no role based control (API completes a findALL() via the end point) . As it stands, anyone can attempt to provision a project in KLAB for instance.

sbathgate (commented):
Overall, this looks good. I have made a couple of notes below that I feel will need to be addressed prior to releasing to PROD. 

Additionally, a very small style adjustment could be made to [SelectInput](https://github.com/bcgov/platform-services-registry/blob/multicluster/web-changes/web/src/components/common/UI/SelectInput.tsx#L12)
```
  <div className={active ? 'active' : ''} style={{flex:1}}>
  ```

Including the `style={{flex:1}}` ensures that the drop down width will scale to match the Ministry. Super trivial but I think it looks nicer than the tiny box it presents as currently.

File: web/src/views/ProfileCreate.tsx (hunk 3 above)
Comment (yhfreeman) [reply web/src/views/ProfileCreate.tsx:76]: ARO is there because it has been included as DB data reference. Let's confirm with the team and remove it if necessary.

The authorization changes can be found in https://github.com/bcgov/platform-services-registry/pull/372.

**Review by sbathgate — APPROVED** [review]

**Review by sbathgate — COMMENTED** [review]
A small observation adjustment to capitalize the first letter of cluster name.

File: web/src/components/profileCreate/CreateFormProject.tsx [hunk 4]
```diff
@@ -31,15 +31,20 @@ interface MinistryItem {
...
   ministry: Array<MinistryItem>;
+  cluster: Array<ClusterItem>;
 }
 
 const CreateFormProject: React.FC<ICreateFormProjectProps> = (props) => {
   const validator = getValidator();
   // @ts-ignore
   const required = (value) => (value ? undefined : 'Required');
-  const { ministry = [] } = props;
+  const { ministry = [], cluster = [] } = props;
```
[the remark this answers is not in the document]
sbathgate: This logic, could also be moved into 'transformDataHelper' in all reality.

yhfreeman:
@sbathgate thanks for your review. I've checked in the code for your suggested changes except uppercasing the cluster names. I can work on that once all the feature priorities are in place.

web/src/views/ProfileCreate.tsx
yhfreeman: > Let's confirm with the team and remove it if necessary.

https://chat.developer.gov.bc.ca/channel/devops-registry?msg=qnakiMHr2vYHXBgiE

PR #373, opened 2021-04-09
Add new github workflow files for api, db, and web

IanKWatts:
For starters, just the web workflow has been replaced.  Once merged into master, it will be possible to manually run the web workflow from the GitHub web UI.  You can select the build config to use at the time the workflow is run.  This will start the build using the specified build config and tag the resulting image with the commit ID and "dev".  The image tag specified in the build config is not important - it is not used by the workflow (so it can be changed from "latest").


sbathgate (approved):
LGTM

PR #380, opened 2021-04-15
Feature/cicd update tags

IanKWatts:


openshift/templates/cicd.yaml
```diff
@@ -66,6 +66,8 @@ objects:
           - create
           - get
           - list
+          - update
+          - delete
```
jleach, line 70: We might be able to get away without delete. Worth a test.

jleach (approved):
Check if we can live without delete. Otherwise looks fine.

Issue #363, opened 2021-03-25
Replace web hard-coded quota values with fetched values from api


yhfreeman:
**Describe the issue**
Currently, the web component has a set of hard-coded quota specs for the t-shirt size small, medium and large. As a dev, in order to make the code base sustainable, I'd like to replace those with fetched values from api.

**Additional context**

- [x] A new route to Registry api so unauthenticated or authenticated users (no preference here) can query quota specs.
- [x] Modify Registry web to adapt api.
- [x] Add unit test and PR review.

**Definition of done**
When the above is done.

yhfreeman:
**Updated on the platform-registry-prod silver cluster: May 03, 2021

PR #383, opened 2021-04-21
Implement quota fetching

AndrewJun97:
Addresses Issue #363.

Fetching Quota sizes from constant values in the WEB has been replaced with fetching from the API


api/src/db/model/quota.ts
```diff
@@ -81,6 +81,48 @@ export default class QuotaModel extends Model {
         // this is intentional (required by Sonarcloud)
     }
 
+    async findQuota(): Promise<any[]> {
```
yhfreeman, line 84: What's the difference you intend between method `findQuota` and base model method `findAll`?

web/src/hooks/useRegistryApi.tsx
```diff
@@ -50,33 +50,43 @@ export default function useRegistryApi() {
   const getMinistry = async (): Promise<AxiosResponse<any>> => {
     if (!axiosInstance.current) {
       throw new Error(errorMsg);
-    } else {
-      return axiosInstance.current.get('ministry');
     }
+    return axiosInstance.current.get('ministry');
+  }
+
+  const getQuota = async (): Promise<AxiosResponse<any>> => {
```
yhfreeman, line 57: Looks like `getQuota` api call is not being used anywhere on the web component. Is there a reason you declared this async function?

yhfreeman (commented):
Thanks for your work Andrew.

FetchQuota and fetchQuotaSizes might not be straightforward for those that are new to this code base. Last time you verbally explained that while they're both for fetching quota t-shirt sizes specifications, one is for the raw version (from db sql) and the other is for represented version used by the frontend.

Perhaps we can think of renaming or adding comments for them.
```
router.get('/', asyncMiddleware(fetchQuota));
router.get('/sizes', asyncMiddleware(fetchQuotaSizes));
```

api/src/db/model/quota.ts
AndrewJun97: Nothing, it looks like I should have used findAll instead.
I've removed the findQuota function now.

web/src/hooks/useRegistryApi.tsx
AndrewJun97: Although it isn't used, I thought it would be nice to have it in there in case it would be needed in the future.
Would you say it's bad practice? Should I remove it?

api/src/db/model/quota.ts
AndrewJun97: Nevermind - findAll() contains the query  `WHERE archived = false` and the db for the quota doesn't have a column for `archived`.
Because of this, don't we need the new `findQuota` function?

AndrewJun97:
> Thanks for your work Andrew.
> 
> FetchQuota and fetchQuotaSizes might not be straightforward for those that are new to this code base. Last time you verbally explained that while they're both for fetching quota t-shirt sizes specifications, one is for the raw version (from db sql) and the other is for represented version used by the frontend.
> 
> Perhaps we can think of renaming or adding comments for them.
> 
> ```
> router.get('/', asyncMiddleware(fetchQuota));
> router.get('/sizes', asyncMiddleware(fetchQuotaSizes));
> ```

Added Commenting for this!

PR #398, opened 2021-05-14
Multi-cluster support feature - Refactor auth and minor fix

yhfreeman:
- refactor around authorization as part of implementation of provisioning test clusters
- minor fix
  - react-table width issue
  - api unit test issue


yhfreeman:
closing due to an issue caused by renaming the local branch, will open another PR with the correct from branch and base branch

PR #399, opened 2021-05-16
Multi-cluster support feature - Refactor auth and minor fix

yhfreeman:
- refactor around authorization as part of implementation of provisioning test clusters

- web component minor fix
  - react-table width issue
  - on profileEdit view, switch to a generic error msg because the frontend does not gracefully handle those requested profile ids that do not exist / are not valid
 
- api component minor fix
  - fix hardcoded provisioning cluster in email
  - rectify a unit test


yhfreeman:
as a result of https://github.com/bcgov/platform-services-registry/pull/398#issuecomment-841786590

PR #409, opened 2021-05-27
Implement an optional prefix config for project licenseplates

sbathgate:
Adjustments to the provisioning system to enable Multi cluster support requires all project namespace license plates to be unique cluster wide. While namespace prefixes are  forced to be unique within the specific environments DB, if a project in two separate environments happened to share a licenseplate value unexpected side effects can be expected due to ArgoCD's current cluster configuration.

This PR enables the Platform Team to set a unique prefix to all license plates within each environment:
`"T" for TEST, "D" for DEV`

As all licenseplates within PROD are restricted to six characters, this will ensure we never have a conflict between environments.


yhfreeman (commented):
The proposed changes would work for preventing TEST and PROD db sharing a prefix (a necessary condition to get argoCD side of things to work).

It might be worth checking the existing patterns like the following
```

**Comment by sbathgate** [comment]
Thanks for the review! From your comment, I am unclear if you are referencing the actual code, as my initial plan was to use the environment variable however TEST is actually populated as production which is why I didn’t utilize that pattern. From what I have gathered adding in a ‘staging’ or ‘test’ environment variable is possible but can cause conflicts. Let me know if I have misunderstood. Otherwise, I will add a comment  to the current pattern as that is definitely a valid observation.

**Comment by yhfreeman** [comment]
@sbathgate now I see the confusion my previous comment may have caused, what I meant was something like
```
export const generateNamespacePrefix = (len: number = 6 | 7): string => {
  let prefix = crypto
    .randomBytes(Math.ceil(len / 2))
    .toString('hex')
    .slice(0, len);

  if (config.get(IS_TEST_SERVICE_INSTANCE)) {
    prefix = 'T' + prefix;
  }
  
  return prefix;
};
```

yhfreeman (approved):
Approve as it works.

PR #412, opened 2021-05-28
Only apply SYNC action name to requests that are already provisioned

### LINKED ISSUE #411: Only apply SYNC action name to requests that are already provisioned
**Describe the issue**
> Found a bug. At least i think it’s a bug. Syncing pending requests replaces the action with sync instead of the action from the original request.
It used to be that we only replaced the action with sync when we were syncing projects that were already provisioned. It’s not a huge issue. Everything else appears to be working. Pending Creation requests are getting the Sync PR body instead of the standard request body so there’s just no information in the PR and they all have the same name.

**Additional context**
https://chat.developer.gov.bc.ca/group/internal-registry-hangout?msg=E3GMwix8bZwSg4X2d

**Definition of done**
When a fix is provided for the above reported issue.

yhfreeman:
Bugfix - Fixes #411

### DIFF
--- a/api/src/controllers/sync.ts
+++ b/api/src/controllers/sync.ts
@@ -139,7 +139,7 @@ export const getProfileBotJsonUnderPending = async (
         throw new Error(errmsg);
       }
 
-      context = await contextForEditing(profileId, true, request.editType, request.editObject);
+      context = await contextForEditing(profileId, request.editType, request.editObject);
     } else {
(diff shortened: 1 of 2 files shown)


sbathgate (approved):
LGTM. Thanks for the quick turn around on this.

yhfreeman:
Bugfix for the recently done multi-cluster feature 5e67dc0

PR #416, opened 2021-05-31
Feature/build for test

IanKWatts:
Adding a second GitHub Action for promotion to Test.  Unlike the original one, which just takes the dev images and tags them for test, the new action will do a build from master for each of the three components that has changed.


sbathgate (approved):
LGTM

PR #433, opened 2021-06-16
Remove unique profile name constraint

sbathgate:
Due to the revised gitops storage organization, teams can now edit project names as they wish. This no longer requires project names to be unique as a result. This PR removes the unique profile name constraint from the database.


jleach (commented):
I though the bots needed this to be true so that they don't confuse the directory layout in GitHub? No longer an issue?

sbathgate:
Correct, with the new directory structure we utilize license-plate names so can edit and duplicate display names as we wish.

jleach (approved):

PR #438, opened 2021-06-22
Cypress Testing

IanKWatts:
The changes here add support for Cypress tests.  This is currently in just the "DEV Build and Tag - WEB" workflow, but could be added to other workflows, of course.

Once we have a service account token to use for posting files to RocketChat, test video files from failed tests could be automatically posted to RocketChat.


sbathgate (approved):
As discussed, we can merge this now so that future Cypress tests are ready to go.

PR #455, opened 2021-07-08
Multiple Technical Leads (Edit)

### LINKED ISSUE #450: Multiple Technical Lead assigned to project
**Describe the issue**
Finalise web adjustments and implement API logic to enable Multiple Technical Leads. The baseline features created by Andrew provide the groundwork for the capability. 

This will involve, refactoring the form data to improve future expansion and better match the modified NATS JSON message. 
```
{
  "project" : {
    "name": "",
    "description": "",
    "prioritySystem": false,
    "busOrgId": "",
    "metadata": [{...}]
  },
  "technical_contacts": [
    {
      "firstName": "po-101",
      "lastName": "po-101",
      "email": "po-101@example.com",
      "rocketchatUsername": null,
      "roleId": 1
    },
    {
      "firstName": "po-101",
      "lastName": "po-101",
      "email": "tl-101@example.com",
      "rocketchatUsername": null,
      "roleId": 2
    },
    {
      "firstName": "po-101",
      "lastName": "po-101",
      "email": "tl-102@example.com",
      "rocketchatUsername": null,
      "roleId": 2
    }
  ]
```

This form data can then be iterated through to make the appropriate API calls.

**Definition of done**
When a user can add a maximum of TWO technical Leads to their project.

sbathgate:
This PR implements the capability to add or edit multiple technical leads to a project during the edit process.

** NOTE: This PR is the depends upon Multiple Technical Leads (Create) **

Fixes #450

### DIFF
--- a/api/src/controllers/profile.ts
+++ b/api/src/controllers/profile.ts
@@ -29,7 +29,6 @@ import { getQuotaSize } from '../libs/profile';
 import { getAllowedQuotaSizes } from '../libs/quota';
 import { requestProfileContactsEdit, requestProfileQuotaSizeEdit, requestProjectProfileCreate } from '../libs/request';
 import shared from '../libs/shared';
-import { validateRequiredFields } from '../libs/utils';
 
 const dm = new DataManager(shared.pgPool);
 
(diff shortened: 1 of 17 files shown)


api/src/controllers/profile.ts
```diff
@@ -74,44 +73,41 @@ export const fetchProfileContacts = async (
 };
 
 export const updateProfileContacts = async (
```
jleach, line 75: This fn might be getting a bit long. Consider breaking it down into smaller fn on your next tech debt cycle.

jleach (approved):

api/src/controllers/profile.ts
sbathgate: 100% agree, Next thing on my docket is extracting alot of the controller logic into separate components as it has definitely gotten quite bloated as the app has grown.

PR #458, opened 2021-07-12
Disable button if user havn't change anything

w8896699:
This pr is just for disabling the button if the user hasn't edit anything compares to last time. 
Will send another pr to use the patch to update only necessary

Ticket link: https://app.zenhub.com/workspaces/platform-experience-5bb7c5ab4b5806bc2beb9d15/issues/bcgov/platform-services-registry/448


web/src/components/profileEdit/ProjectCardEdit.tsx
```diff
@@ -49,6 +49,33 @@ interface MinistryItem {
   code: string;
 }
 
+export const renderSubmitButton = (hasPendingEdit: any, isProvisioned: any, pristine: boolean) => {
```
sbathgate, line 52: Good decision to extract this into a separate component. 

nit pick: I wonder if it could be extracted to its own file, maybe under utils, to avoid nested common logic within a single file.

sbathgate (approved):
Other than the minor nitpick this looks good.

web/src/components/profileEdit/ProjectCardEdit.tsx
w8896699: updated

sbathgate (approved):
LGTM

PR #459, opened 2021-07-12
Disable edit logic for Quota edit

w8896699:
Ticket: https://app.zenhub.com/workspaces/platform-experience-5bb7c5ab4b5806bc2beb9d15/issues/bcgov/platform-services-registry/403
remove the logic to hide the ‘edit’ button if pending edits, that way users can still get to the edit page even if there is a pending edit.

### DIFF
--- a/web/src/views/ProfileEdit.tsx
+++ b/web/src/views/ProfileEdit.tsx
@@ -226,11 +226,12 @@ const ProfileEdit: React.FC = (props: any) => {
                     <Text as="h3" color={theme.colors.contrast} mx={2}>
                       {c.title}
                     </Text>
-                    {!profileState.hasPendingEdit && profileState.isProvisioned && (
-                      <RouterLink className="misc-class-m-dropdown-link" to={c.href}>
-                        <Icon hover color="contrast" name="edit" width={1.5} height={1.5} />
-                      </RouterLink>
-                    )}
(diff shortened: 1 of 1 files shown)


web/src/views/ProfileEdit.tsx
```diff
@@ -226,7 +226,8 @@ const ProfileEdit: React.FC = (props: any) => {
                     <Text as="h3" color={theme.colors.contrast} mx={2}>
                       {c.title}
                     </Text>
-                    {!profileState.hasPendingEdit && profileState.isProvisioned && (
+                    {((!profileState.hasPendingEdit && profileState.isProvisioned) ||
```
sbathgate, line 229: With the revisions in #458 I think you could remove this logic entirely. Eventually, it would be good to `disable` fields from being able to be edited if their is a pending edit.

sbathgate (changes requested):
As commented above ^

web/src/views/ProfileEdit.tsx
w8896699: updated

w8896699:
> As commented above ^

Thanks for reminding!, I have removed this logic, But I suggest that this pr will be merge after https://github.com/bcgov/platform-services-registry/pull/458

sbathgate (approved):
LGTM

PR #460, opened 2021-07-12
Add a link to the quota increase document

w8896699:
Ticket : https://app.zenhub.com/workspaces/platform-experience-5bb7c5ab4b5806bc2beb9d15/issues/bcgov/platform-services-registry/355

### DIFF
--- a/web/src/components/profileEdit/QuotaCardEdit.tsx
+++ b/web/src/components/profileEdit/QuotaCardEdit.tsx
@@ -256,7 +256,7 @@ const QuotaCardEdit: React.FC<IQuotaCardEditProps> = (props) => {
               flexDirection="column"
               sx={{ borderRadius: ' 10px', alignItems: 'center' }}
             >
-              <FormTitle>Upgrade Quota</FormTitle>
+              <FormTitle>Upgrade/Downgrade Quota</FormTitle>
               <Flex mt={3}>
                 <Label m="auto" width={3 / 4}>
                   <Text as="h3" fontSize="16px" my={0} lineHeight="normal">
--- a/web/src/components/profileEdit/QuotaCardEdit.tsx
+++ b/web/src/components/profileEdit/QuotaCardEdit.tsx
@@ -16,6 +16,7 @@
 
 import { Label } from '@rebass/forms';
 import React, { useEffect, useState } from 'react';
+import styled from '@emotion/styled';
 import { Redirect } from 'react-router-dom';
 import { Text } from 'rebass';
 import { PROFILE_EDIT_VIEW_NAMES, ROUTE_PATHS } from '../../constants';
@@ -36,6 +37,17 @@ interface IQuotaCardEditProps {
   hasPendingEdit: boolean;
 }
 
+const StyledExternalLink = styled.a`
+  color: #003366;
+  font-weight: 600;
+  :visited: {
+    color: #003366;
+  }
+`;
+
+const StyledCheckbox = styled.input`
+  margin: 10px 10px 10px 0;
+`;
 const QuotaCardEdit: React.FC<IQuotaCardEditProps> = (props) => {
   const {
     quotaDetails: { licensePlate = '', quotaSize = '', quotaOptions = [] },
@@ -51,9 +63,16 @@ const QuotaCardEdit: React.FC<IQuotaCardEditProps> = (props) => {
   const [goBackToProfileEditable, setGoBackToProfileEditable] = useState<boolean>(false);
   const [selectedSize, setSelectedSize] = useState<any>('');
   const [specs, setSpecs] = useState<any>([]);
+  const [validateSelection, setValidateSelection] = useState<boolean>(false);
+  const [isReadInstruction, SetIsReadInstruction] = useState<boolean>(false);
 
   const handleChange = (event: any) => {
-    setSelectedSize(event.target.value);
+    if (event.target.value === 'select') {
+      setValidateSelection(false);
+    } else {
+      setValidateSelection(true);
+      setSelectedSize(event.target.value);
+    }
   };
 
   const handleSubmit = async () => {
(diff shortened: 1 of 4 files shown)


web/src/components/profileEdit/QuotaCardEdit.tsx
```diff
@@ -128,8 +147,28 @@ const QuotaCardEdit: React.FC<IQuotaCardEditProps> = (props) => {
         {specs.storageNums[1]} overall storage with {specs.storageNums[2]} for backup storage
       </Text>
       <br />
+
+      <p>
```
sbathgate, line 151: During review this portion of code seems to cause an issue with the layout, as seen in the attached screen shot. The expected result would be for the Quota information card to populate ~60% of the window with the text blurb on the right side as seen in the second screenshot.

Additionally, the use of flat HTML has typically avoided. You will see in this page an example on line #144 that demonstrates the use of <Text>. I will note, this page could definitely be subject to a more thorough UI/UX enhancement if you wished to rewrite the component to be more user friendly. I can draft a ticket if that were the case.
```suggestion
      <Text as='p'>
```


Screen Shot 2021-07-12 at 4 27 31 PM
Screen Shot 2021-07-12 at 4 29 30 PM

web/src/components/profileEdit/QuotaCardEdit.tsx
```diff
@@ -51,9 +63,16 @@ const QuotaCardEdit: React.FC<IQuotaCardEditProps> = (props) => {
   const [goBackToProfileEditable, setGoBackToProfileEditable] = useState<boolean>(false);
   const [selectedSize, setSelectedSize] = useState<any>('');
   const [specs, setSpecs] = useState<any>([]);
```
sbathgate, line 65: I also have noticed a bug that only seems to appear in this branch locally, if you manually goto a page that should be hidden (pending edit or not provisioned yet) I receive the following error:
```
TypeError: specs is undefined
QuotaCardEdit
src/components/profileEdit/QuotaCardEdit.tsx:122

  119 |   );
  120 | }
  121 | 
> 122 | if (specs.length === 0) {
      | ^  123 |   return null;
  124 | }
  125 | return (
  ```
  
  It would be good to see if you could replicate this or if it is just a local issue for me.
  
  To reproduce, create a new project and goto the project overview page. Manually update the URL from `.../:projectId/overview` to `.../:projectId/quota` and see if you receive a similar alarm.

File: web/src/components/profileEdit/QuotaCardEdit.tsx (hunk 1 above)
Comment (w8896699) [reply web/src/components/profileEdit/QuotaCardEdit.tsx]: I have updated using styled component and add a style to it.

File: web/src/components/profileEdit/QuotaCardEdit.tsx (hunk 2 above)
Comment (w8896699) [reply web/src/components/profileEdit/QuotaCardEdit.tsx:69]: I tried on my localhost and docker and l can't reproduce this issue.

Comment (sbathgate) [reply web/src/components/profileEdit/QuotaCardEdit.tsx:69]: Interesting, we can take a look when this is on DEV.

**Review by sbathgate — CHANGES_REQUESTED** [review]
I have made a couple of notes on the ProjectEdit.tsx adjustments. I feel part of the problem stems from how the QuotaEdit card has a varying structure to the rest of the application. 

I feel this would be a good opportunity to rewrite this component to more closely replicate the structure of ProjectCardEdit and ContactCardEdit.
I have mocked up what I think would be a workable solution that would make this component match the rest of the webapp.

All of the fields in the white box will be disabled permanently and are strictly for information purposes.

Review the 'Review' Modal on the dashboard for an example of what this page could look like. Note: this is a component based page, the modal just has content that resembles what I have in mind.

This mock up can be found in the Registry Miro board too. Don't hesitate to ping me if you have any questions.

image

File: web/src/components/profileEdit/QuotaCardEdit.tsx [hunk 3]
```diff
@@ -100,62 +181,152 @@ const QuotaCardEdit: React.FC<IQuotaCardEditProps> = (props) => {
...
+                    ),
+                  )}
+                </Flex>
+              </Box>
+            ),
+          );
+
+          const QuotaChangeComponent = (
+            <Flex
+              backgroundColor={theme.colors.grey}
```
sbathgate, line 252: I feel this grey is far too dark for the purposes of a background colour in this circumstance. Similar to the request approval modal, something like `#eeeeee` is much kinder on the eyes.

web/src/components/profileEdit/QuotaCardEdit.tsx
```diff
@@ -100,62 +181,152 @@ const QuotaCardEdit: React.FC<IQuotaCardEditProps> = (props) => {
...
+              paddingX={4}
+              paddingY={3}
```
sbathgate, line 254: Similar to other notes, matching the style used throughout is definitely valuable.
```suggestion
              px={4}
              py={3}
```

web/src/components/profileEdit/QuotaCardEdit.tsx
```diff
@@ -36,7 +41,18 @@ interface IQuotaCardEditProps {
...
+  max-width: 35vw;
+  padding: 0px;
+  @media only screen and (max-width: 680px) {
+    max-width: 90vw;
+  }
+`;
+
 const QuotaCardEdit: React.FC<IQuotaCardEditProps> = (props) => {
+  // @ts-ignore
+  const required = (value) => (value ? undefined : 'Required');
```
sbathgate, line 54: I recognize this was likely copied from the CreateFormRequest card, but that `// @ts-ignore` can be removed and the circumstance resolved with the following change.
```suggestion
  const required = (value: boolean) => (value ? undefined : 'Required');
```

sbathgate (changes requested):
I am going to revisit this PR, as the logic used at the base seems overly complex for the functionality this feature requires. Overall though great improvement over the first draft. There are a few notes that can be actioned/considered in the meantime

web/src/components/profileEdit/QuotaCardEdit.tsx
```diff
@@ -81,11 +159,14 @@ const QuotaCardEdit: React.FC<IQuotaCardEditProps> = (props) => {
   };
 
   useEffect(() => {
-    async function getQuotaSizes() {
-      const quotaSizes = await api.getQuotaSizes();
-      setSpecs(quotaSizes.data.filter((size: any) => size.name === quotaSize).pop());
-    }
-    getQuotaSizes();
+    (async () => {
```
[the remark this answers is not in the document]
w8896699: I think the function's purpose can be read from line 164: api.getQuotaSizes(). 

It's just I think this function is not been reused anywhere. My main purpose of changing here is just to use try and catch catch errors. It is good practice to always use try and catch on to any aync call.

web/src/components/profileEdit/QuotaCardEdit.tsx
```diff
@@ -49,22 +65,84 @@ const QuotaCardEdit: React.FC<IQuotaCardEditProps> = (props) => {
...
 
   const [goBackToProfileEditable, setGoBackToProfileEditable] = useState<boolean>(false);
-  const [selectedSize, setSelectedSize] = useState<any>('');
   const [specs, setSpecs] = useState<any>([]);
+  const [applyingQuotaSpecs, setApplyingQuotaSpecs] = useState<any>([]);
+  const [quotaSizes, setQuotaSizes] = useState<any>({});
 
-  const handleChange = (event: any) => {
-    setSelectedSize(event.target.value);
+  const QUOTAINFORMATIONCONSTANTS: any = {
```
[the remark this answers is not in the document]
w8896699: I don't think it can be placed in constants.ts file, as some of the value is getting in this component

w8896699:
I have updated code  and replied comment

sbathgate (approved):
There is only one final thing I observe with this PR and that is the 'Request Quota Change' button not scaling quite correctly. This could be carry over from before though, but something that should be addressed. This can be tackled in a separate PR to enable this one to be merged.
Screen Shot 2021-08-23 at 12 49 17 PM

PR #466, opened 2021-07-19
Add RocketChat notifications on failed build tests

IanKWatts:


.github/workflows/dev-web.yml
```diff
@@ -61,20 +60,50 @@ jobs:
...
           CI: "true"
-      - name: Cypress test
+
+      - name: Check build result
+        if: ${{ failure() }}
+        run: echo "Error in build step"; exit 1
+
+      - name: npm start
+        run: |
+          npm start
```
sbathgate, line 77: I am not sure if it would be worthwhile defining the port explicitly here? Just where it seems Cypress relies on it being :3000.
```suggestion
          PORT=3000 npm start
```

sbathgate (approved):
Other than the Port comment, LGTM.

.github/workflows/dev-web.yml
IanKWatts: I have added the port to the npm start line.

PR #480, opened 2021-08-06
Update dashboard api logic to include getDashboardProjects service

sbathgate:
This PR introduces the API logic necessary to add multiple TL's to the dashboard view. This includes an overhaul on the API call used to fetch profiles.

This is the root of the Multiple-TL's PR's, a WEB PR will follow with the UI adjustments that utilize this new API endpoint.

As the application has begun to expand, it is clear the controllers are beginning to get fairly large with business logic. This PR also introduces the new design pattern of a Service that correlates with the controller. An issue will be drafted to commit development effort to transition the current controllers to this pattern.


api/__tests__/__snapshots__/project-profile.spec.ts.snap
```diff
@@ -319,7 +319,14 @@ Array [
   Array [
     Object {
       "text": "
-      SELECT DISTINCT(profile.id), profile.* FROM profile
+      SELECT 
+        DISTINCT(profile.id),
```
jleach, line 323: Do you need DISTINCT here? I think `id` is the pk and is guaranteed to be unique.

api/src/db/model/profile.ts
```diff
@@ -293,7 +293,14 @@ export default class ProfileModel extends Model {
   async findProfilesByUserIdOrEmail(id: number, email: string): Promise<any> {
     const query = {
       text: `
-      SELECT DISTINCT(profile.id), profile.* FROM ${this.table}
+      SELECT 
+        DISTINCT(profile.id),
```
jleach, line 297: DISTINCT?

jleach (approved):

PR #489, opened 2021-08-23
Gold dr features

sbathgate:


api/src/libs/messaging.ts
```diff
@@ -58,12 +105,13 @@ const updateEmailContent = async (
...
       projectDescription : profile.description,
       projectMinistry: profile.busOrgId,
-      setCluster: profile.primaryClusterName,
+      setCluster: clusters.map(cluster => cluster.displayName).toString(),
       licensePlate: `${profile.namespacePrefix}`,
       requestType: ( request.editType ? 'Project Edit': 'New Project'),
       quotaSize: ( request.editType === 'quotaSize' ? request.editObject.quota : 'Small'),
       editType: ( request.editType === 'quotaSize' ? 'Quota' : ''),
       humanActionComment,
+      consoleButtons: clusters.map(cluster => consoleButton(cluster.name)).toString(),
```
jleach, line 114: Same here.

api/src/libs/profile.ts
```diff
@@ -29,12 +29,19 @@ const { NamespaceModel, ClusterModel, ProfileModel, ContactModel } = dm;
...
-    if (!primaryCluster.id || !profile.id) {
-      throw new Error('Unable to get primary cluster id or profile id');
+    const clusters = await getClusters(profile)
+    for (const cluster of clusters) {
+      const isClusterProvisioned = await NamespaceModel.getProjectSetProvisionStatus(
+        Number(profile.id),
+        Number(cluster.id)
+      );
+
+      if (!isClusterProvisioned){
```
jleach, line 39: Can you just `return isClusterProvisioned`? or `return isClusterProvisioned ? true : false`.

api/src/services/profile.ts
```diff
@@ -40,9 +40,17 @@ export const fetchAllDashboardProjects = async (userDetails: any): Promise<any>
       return displayName;
     }
 
+    const extractGoldDR = (clusters) => {
+      if (clusters.length === GOLD_QUORUM_COUNT) {
+        return clusters.filter(cluster => cluster.name === 'gold')
```
jleach, line 45: `gold` might be better done as a constant.

jleach (approved):
Check the `map` and return value comment. If they are OK with you then they are OK with me.

PR #502, opened 2021-08-27
Remove technical lead from contact

w8896699:
https://app.zenhub.com/workspaces/platform-experience-5bb7c5ab4b5806bc2beb9d15/issues/bcgov/platform-services-registry/477


api/src/controllers/profile.ts
```diff
@@ -89,18 +90,21 @@ export const updateProfileContacts = async (
...
     const provisionerContactEdit: boolean[] = [];
+    const removeExistingContact = currentContacts.filter(comparerContact(contacts));
+    const addNewContact = contacts.filter(comparerContact(currentContacts));
+    const AddOrRemoveContact = removeExistingContact.concat(addNewContact);
 
+    if(AddOrRemoveContact.length !== 0){
+      provisionerContactEdit.push(deletedOrNewContact)
+    }
     // 1. Check for provisioner related changes
     contacts.forEach((contact: Contact): void => {
```
sbathgate, line 101: Nit pick, move the comment to line 96 (include appropriate spacing

Also, would it be worth adding this as an `else` to the if condition above as it would save the additional processing if you know it is already a provisioner edit request.

api/src/db/model/profile.ts
```diff
@@ -249,6 +249,29 @@ export default class ProfileModel extends Model {
...
+
+    try {
+      query.values = [
+        contactId,
+        profileId,
+      ];
+      const results = await this.runQuery(query);
+      return results.pop();
+    } catch (err) {
+      const message = `Unable to remove contact ${contactId} from contact profile ${profileId}`;
```
sbathgate, line 269:
```suggestion
      const message = `Unable to remove contact ${contactId} from profile ${profileId}`;
```

api/src/db/model/profile.ts
```diff
@@ -249,6 +249,29 @@ export default class ProfileModel extends Model {
       throw err;
     }
   }
+  
+  async removeContactFromProfile(profileId: number, contactId: number): Promise<void> {
+    const values: any[] = [];
+    const table = 'profile_contact';
+    const query = {
+      text: `DELETE FROM ${table} WHERE contact_id = $1 AND profile_id = $2 RETURNING *;`,
```
sbathgate, line 257: This is something we need to clean up in the API at somepoint. It would be advantageous to pick a formatting style and stick to it. ${table} uses the variable, where $1 uses the values syntax. I recall @jleach had looked into a best practice here previously.

api/src/db/utils.ts
```diff
@@ -43,3 +43,9 @@ export const generateNamespacePrefix = (len: number = 6): string => {
     .toString('hex')
     .slice(0, len);
 };
+
+export const comparerContact = (otherArray) =>{
```
sbathgate, line 47: This could use a bit of a clean up, it is a little tough to understand what is going on with it.
title, spacing, missing type, missing ;, variables are unclear (if the method is called compareContacts then the internal variables should reflect this), 
```suggestion
export const compareContacts = (otherArray) => {
```

api/src/libs/request.ts
```diff
@@ -125,7 +146,14 @@ export const processProfileContactsEdit = async (request: Request): Promise<void
...
 
-        // TODO(sb): implement functionality to delete a contact if a project goes from 2 TL's -> 1 TL.
+        // functionality to delete a contact if a project goes from 2 TL's -> 1 TL.
+        const removeExistingContact = currentContacts.filter(comparerContact(contacts));
+        //remove contact
+      if(removeExistingContact[0]) {
+        const removeExistingContactID = removeExistingContact[0].id;
+        await ProfileModel.removeContactFromProfile(Number(request.profileId), Number(removeExistingContactID))
+        await ContactModel.delete(Number(removeExistingContactID))
+      }
```
sbathgate, line 156: indentation
```suggestion
        if(removeExistingContact[0]) {
          const removeExistingContactID = removeExistingContact[0].id;
          await ProfileModel.removeContactFromProfile(Number(request.profileId), Number(removeExistingContactID))
          await ContactModel.delete(Number(removeExistingContactID))
        }
```

api/src/libs/request.ts
```diff
@@ -125,7 +146,14 @@ export const processProfileContactsEdit = async (request: Request): Promise<void
             }
         }
 
-        // TODO(sb): implement functionality to delete a contact if a project goes from 2 TL's -> 1 TL.
+        // functionality to delete a contact if a project goes from 2 TL's -> 1 TL.
+        const removeExistingContact = currentContacts.filter(comparerContact(contacts));
+        //remove contact
+      if(removeExistingContact[0]) {
```
sbathgate, line 152: Magic number. I believe it is better practice to use something like `.pop()` to extract the first value.

sbathgate (changes requested):
Few things to clean up. But overall this will definitely be good to get included.

api/src/db/model/profile.ts
w8896699: It is actually removed from the contact profile instead of the profile

api/src/db/utils.ts
w8896699: updated to use lodash instead of self-writing function

api/src/db/model/profile.ts
w8896699: I will create a separate ticket for those.

api/src/libs/request.ts
w8896699: updated to use map loop instead of if loop

api/src/controllers/profile.ts
```diff
@@ -18,6 +18,7 @@
 
 import { errorWithCode, logger } from '@bcgov/common-nodejs-utils';
 import { Response } from 'express';
+import * as _ from "lodash";
```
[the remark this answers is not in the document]
w8896699: It is hard to explain what I did in that original request function, therefore for readability consideration, I decide to just use library.

api/src/db/utils.ts
```diff
@@ -42,4 +42,4 @@ export const generateNamespacePrefix = (len: number = 6): string => {
     .randomBytes(Math.ceil(len / 2))
     .toString('hex')
     .slice(0, len);
-};
+};
```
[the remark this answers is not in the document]
w8896699: I can't see this line in my editor 😢

sbathgate (approved):
This PR looks good now, just a note for the future though, it is important to use more valueable commits than "Resolve comment" that doesn't tell the person anything about this. Take a read through this [link](https://chris.beams.io/posts/git-commit/) but the gist is:
> A properly formed Git commit subject line should always be able to complete the following sentence:

    "If applied, this commit will [...your subject line here]"
 
Technically "Resolve comment" would work but it provides zero value.

PR #512, opened 2021-09-09
PR github action

w8896699:


web/src/__tests__/Dashboard.test.tsx
```diff
@@ -87,5 +87,5 @@ function renderDashboard() {
 test('<Dashboard / > Table view should render', async () => {
   const { container } = renderDashboard();
 
-  await waitFor(() => expect(container).toMatchSnapshot());
+  await waitFor(() => expect(container).toBeInTheDocument());
```
sbathgate, line 90: Further clarification here would be beneficial. This seems like it makes the point of the test entirely redundant, as it now just verifies that it renders something.

.github/workflows/promote_to_branch.yml
```diff
@@ -0,0 +1,78 @@
...
+          node-version: "12"
+      - name: Install dependencies
+        run: npm ci
+      - name: Unit tests w/ coverage
+        run: npm run test:coverage
+      - name: LINTing
+        run: npm run test:lint
+      - name: OpenAPI Schema
+        run: npm run test:schema
+      - name: Upload coverage report
```
sbathgate, line 34: We haven't been using this coverage report system, is there a reason you have chosen to include it?

.github/workflows/promote_to_branch.yml
```diff
@@ -0,0 +1,78 @@
...
+        working-directory: ./web
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "12"
+      - name: Install dependencies
+        run: npm ci
+      - name: test
+        run: ls src/__tests__/__snapshots__/
```
sbathgate, line 64: I am unsure of the value this step adds. Any particular reason for its inclusion?

sbathgate (commented):
I am wondering if it is worth only running these on pushes to PR's vs just any branch as we could hammer the registry hangouts channel with spam from every commit to remote branches.

web/src/__tests__/Dashboard.test.tsx
w8896699: toMatchSnapshot somehow doesn't work for this component after I merge the table sort or, so I use toBeInTheDocument for now.

web/src/__tests__/Dashboard.test.tsx
```diff
@@ -87,5 +87,5 @@ function renderDashboard() {
 test('<Dashboard / > Table view should render', async () => {
```
patricksimonian, line 87:
```suggestion
test.skip('<Dashboard / > Table view should render', async () => {
```

patricksimonian (approved):

PR #539, opened 2021-10-22
Fed 1639 snapshot quotas

w8896699:


api/src/libs/quota.ts
```diff
@@ -49,14 +49,15 @@ export const getAllowedQuotaSizes = (
...
+    Object.keys(currentQuotaSize).forEach(
+      // eslint-disable-next-line no-return-assign
+      (key) =>
+        (availableQuotaOptions[key] = getAllowQuotaForEachResource(
           currentQuotaSize[key]
-        );
-      }
-    }
+        ))
+    );
```
patricksimonian, line 60: this is not terrible but this can be expressed in a `.reduce` fyi

patricksimonian (approved):
your code is getting much better 🔥

PR #574, opened 2021-11-18
Add ministry id to NATS context message

sbathgate:
This PR introduces the Ministry ID into the NATS Context. Which is necessary for the Warden Operator and ArgoCD Shared so we don’t have to manage the list of namespaces attached to an AppProject manually.


w8896699 (approved):
LGTM

Issue #586, opened 2022-01-05
Remove redundant dropdown in desktop view


w8896699:
**Describe the issue**
In the registry dashboard, the `create` button will give a drop-down where only have one available option, that can be removed only in desktop view

**Additional context**
Gordon's first practice ticket.

**Definition of done**
Click create button will directly go to create the page. 
Gordon successfully create a pr and merge into master

PR #590, opened 2022-01-07
Fix redundant dropdown menu in desktop mode #586

qrdn:
Fix for redundant dropdown menus.

Uses "useHistory" React hook from 'react-router-dom'.
Newer versions of react-router seem to use "useNavigate" instead.

Dropdown UI elements not removed.


web/src/components/common/layout/Header.tsx
```diff
@@ -81,20 +79,18 @@ interface INavProps {
 const Nav: React.FC<INavProps> = (props) => {
   const { isAuthenticated, handleDDMobile, isDDMobileOpen, dirs } = props;
 
-  const { ref, isComponentVisible, setIsComponentVisible } = useComponentVisible(false);
+  const history = useHistory();
 
-  const handleDDDesktop = () => {
-    setIsComponentVisible(!isComponentVisible);
+  const handleCreateDesktop = () => {
+    //  newer versions of react uses useNavigate instead of useHistory from 'react-router-dom'
```
w8896699, line 85: Leave a TODO and your name at this line so we can come back to address it later when you have more experience.

w8896699 (changes requested):

w8896699 (approved):
Looks good to me! 🔥

PR #594, opened 2022-01-12
Fix CSP

w8896699:
https://app.zenhub.com/workspaces/platform-experience-5bb7c5ab4b5806bc2beb9d15/issues/bcgov/platform-services-registry/540

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,7 +1,6 @@
 {
   "version": "0.2.0",
   "configurations": [
-  
     {
       "name": "Attach to API (docker)",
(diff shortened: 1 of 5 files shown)


okanji (commented):
Looks good @w8896699! Feel free to merge this in, my only question is that is it necessary to commit the changes to ```.vscode/launch.json```?

w8896699:
It’s changed because we updated our prettier rule recently and I checked that file when I am working on this pr, so prettier automatically update its linting😂. I will didn’t noticed that when I commit the change, I will drop that file!

PR #599, opened 2022-01-14
Update zape scan site

w8896699:
Change Scanning URL from dev to prod URL
It makes more sense for our zape security tools to scan our production URL instead of the dev environment.


okanji:
@w8896699 , can you provide a short description for this PR please including why the target UIRL is being changed? Thanks!

Issue #555, opened 2021-11-08
Create read-only admin account in Registry for 7-7000 agents


mitovskaol:
To effectively assist teams calling 7-7000 for application support after hours, it would be helpful for 7-7000 agents to have read-only access to the Project Registry so that they can look up who the Technical Lead for the app in trouble as sometimes the caller may not know this information.

DoD:
- [x] An account created in the Registry that has read-only access to all projects 
- [ ] The account details provided to @sbathgate to include in the 7-7000 troubleshooting script

okanji:
We can now assign 7-7000 users the **read-only-administrator** role in keycloak to prohibit them from making any admin write changes while still having read access. This is now in production. @mitovskaol will take care of assigning the 7-7000 agents to this new role.

PR #607, opened 2022-01-31
Created a "read only administrator" authorization role 

okanji:
I have added support for a new user role **read_only_administrator** that has the permissions of a regular user and only the read permissions of an administrator. A **read_only_administrator** can not approve or reject a project but can view its information. It includes changes to the `authorization` script and authorization checks in various controller functions that require admin authorization as well as corresponding tests and changes to the web ui. For more background, see ticket #555 

##

These changes will allow 7-7000 agents to be able to view administrator data without being able to reject and review projects. 

##

Currently, **administrator** is the only role assigned to a human user. The **administrator** is associated with the`EditAll` authorization flag. A regular user has no associated authorization flags. 

I have added an additional flag `ReadAll` and added it to the **administrator** role in addition to the existing `EditAll` flag. The **read_only_administrator** is also assigned the `ReadAll` flag but not `EditAll`. The  **read_only_administrator** is also assigned the `ProvisionOnTestCluster` flag which is also assigned to the **administrator** role as it authorizes some fetch request functions but does not authorize any update functions (and so it does not give the **read_only_administrator** any write access).

##

Tests have been added for corresponding functions that use the newly created `ReadAll` authorization flag.

## 

I have disabled the review button in the web application for a **read_only_administrator**. I tried to delete the whole _response_ cell in the table but doing this caused the link in the name cell to become disabled. I am not sure why this is the case or how the two are connected and I chose not to dig deeper into this issue as it is purely an aesthetic issue and I feel that I have spent enough time on this PR already, but I can look further into it if needed.


web/src/views/Dashboard.tsx
```diff
@@ -74,9 +74,10 @@ const Dashboard: React.FC = () => {
     verifyProjects();
   }, 1000 * 30);
 
+
   return (
     <>
-      {userRoles.includes('administrator') && <ProjectRequests profileDetails={profileDetails} />}
```
w8896699, line 79: I am not sure if we want to let them see all those requests, we can check this with Olena!

web/src/components/dashboard/ProjectRequests.tsx
```diff
@@ -74,20 +74,22 @@ const ProjectRequests: React.FC<any> = (props) => {
...
-          </Box>
-        ),
+        Cell: ({ row: { values } }: any) =>
+          isAdminUser && (
+            <Box>
+              <button
+                type="button"
+                value={values.profileId}
+                onClick={() => {
+                  console.log('Click');
```
w8896699, line 84: please remove console.log( I suggest we can add no console rule later in our eslint rule)

w8896699 (changes requested):
Looks good to me, just a little change.

web/src/components/dashboard/ProjectRequests.tsx
okanji: Yes thank you for catching that!

okanji:
@w8896699 re your comment about not being sure if the read only admin should be able to see all the requests, I am under the understanding that the read only admin should be able to see exactly what a regular admin sees. If you can be more specific in what you think that the read only admin I will relay that to Olena to get some direction

w8896699 (approved):
I agree with your comment above. Looks good to me now! I can show you how our pipeline works when you want to merge this pr.

PR #611, opened 2022-02-02
klab tool quota downgrade

w8896699:
https://app.zenhub.com/workspaces/platform-experience-5bb7c5ab4b5806bc2beb9d15/issues/bcdevops/developer-experience/2205

For testing purposes, we first downgrade tool namespace quota in klab/clab first

This pr also solves a duplicate key problem where sometimes a project has the same first and second Technical lead.

### DIFF
--- a/api/src/libs/profile.ts
+++ b/api/src/libs/profile.ts
@@ -182,6 +182,7 @@ export const getQuotaSize = async (
   try {
     const clusters: Cluster[] = await getClusters(profile);
     const promises: any = [];
+
     clusters.forEach((cluster: Cluster) => {
       if (!profile.id || !cluster.id) {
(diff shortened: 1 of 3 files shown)


api/src/libs/profile.ts
```diff
@@ -211,10 +206,6 @@ export const getQuotaSize = async (
       hasSameQuotaSizesForAllClusters = profileQuotaSizes.every((element) =>
         isEqual(element, profileQuotaSizes[0])
       );
-
-      hasSameQuotaSizesForAllClusters = compareNameSpaceQuotaSize(
```
w8896699, line 215: This is redundant code that I forget to remove from before.

okanji (approved):

PR #621, opened 2022-02-17
Technical Lead fields now say (Primary) or (Secondary) as appropriate

ShovelHand:
In create as well as edit contexts.

Pretty darn straightforward.

### DIFF
--- a/web/src/components/profileCreate/CreateFormTL.tsx
+++ b/web/src/components/profileCreate/CreateFormTL.tsx
@@ -45,7 +45,9 @@ const CreateFormTL: React.FC = () => {
             {fields.map((name, index) => (
               <div key={name}>
                 <Flex flexDirection="row">
-                  <FormTitle style={{ margin: '14px 0 5px 0' }}>Technical Lead {index == 0 ? '(Primary)' : '(Secondary)'}</FormTitle>
+                  <FormTitle style={{ margin: '14px 0 5px 0' }}>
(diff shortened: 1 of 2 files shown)


web/src/components/profileCreate/CreateFormTL.tsx
```diff
@@ -45,7 +45,7 @@ const CreateFormTL: React.FC = () => {
             {fields.map((name, index) => (
               <div key={name}>
                 <Flex flexDirection="row">
-                  <FormTitle style={{ margin: '14px 0 5px 0' }}>Technical Lead</FormTitle>
+                  <FormTitle style={{ margin: '14px 0 5px 0' }}>Technical Lead {index == 0 ? '(Primary)' : '(Secondary)'}</FormTitle>
```
w8896699, line 48: Normally we use `===` in javascript to make sure they fully match.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality

w8896699: You will need to update snapshot and make sure lints are fine by running:
```
in ../web run:
npm run test -- -u
Fir just as an FYI. the --u will update the snapshots captured during testing.
npm run lint -- --quiet -f codeframe --fix
npm run format -- --write
```

w8896699 (changes requested):

PR #628, opened 2022-02-24
Error page

ShovelHand:
I think it would be a good idea to add to this information for contacting technical support, whoever that may be. I'm not sure what details are appropriate for that. 
So far this is only the error page component that users should see when a React error occurs. There is not yet any actual error handling to direct users to this page. 
I see there are some tests that failed when I made my first commits two days ago. I don't think theses are still issues.

image


web/src/views/ErrorPage.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+const ErrorPage: React.FC<Props> = ({errorMessage}) => {
+  return (
+    <div>
+      <h1>React Error</h1>
+      Something seems to have gone wrong...
+      <div id="errorPageMessage">
+        Error message: {errorMessage}
+      </div>
+      <Link to="/">Go back to the home page</Link>
```
w8896699, line 32: try to find a better solution instead of using Link

https://www.freecodecamp.org/news/react-router-tutorial/

web/src/views/ErrorPage.tsx
```diff
@@ -0,0 +1,37 @@
...
+interface Props {
+    errorMessage: string
+}
+
+const ErrorPage: React.FC<Props> = ({errorMessage}) => {
+  return (
+    <div>
+      <h1>React Error</h1>
+      Something seems to have gone wrong...
+      <div id="errorPageMessage">
```
w8896699, line 29: We  use rebass across the project, try to follow the pattern

w8896699 (changes requested):
Always remember to run lint and unit test before push

web/src/views/ErrorPage.tsx
ShovelHand: Please be more specific. I have no idea what we're actually trying to address here. I used NotFound.tsx as an example for this view as I figured they were reasonably close in terms of function.

okanji:
Looks good @ShovelHand, a few changes that I have requested above. Also, can we align the text "Error" with the text below, they seem to be horizontally misaligned. 

As for Billy's review comment about using ** rebass**: it is a styling library that can be used to make styling more structured and on the component level. There are multiple alternatives to it and it is common to use such a package in a modern React web app. That being said, we probably will not use **rebass** in the re write so its actually nice that this component does not use it as it will be easy to use it in the re write.

Thanks!

okanji:
Hey @ShovelHand, thanks for making those changes! Could I ask that you upload a screenshot of the error page like the one above so I can take a look? Thank you!

PR #639, opened 2022-03-01
Removed all TODOs in web and converted to Zenhub Issues

qrdn:


api/src/libs/fulfillment.ts
```diff
@@ -405,6 +405,7 @@ export const sendNatsMessage = async (
   try {
     const nc = shared.nats;
     const { natsSubject, natsContext } = natsMessage;
+    console.log(natsContext);
```
w8896699, line 408: remove console.log

w8896699 (changes requested):

PR #646, opened 2022-03-08
Fix/whitespace trimming

qrdn:
This trims the whitespace of the project description.
I placed the code in a function used previously for a double quote issue.

The whitespace will still occur in the current logger in "sendNatsMessage", as the "replaceForDescription" is called after the logger is called.


w8896699 (approved):
Looks good to me.

PR #650, opened 2022-03-09
Web env vars

okanji:
Currently, **.env** variables are only being exposed to the API and not the WEB in development, this change allows the web app to source **.env** variables


w8896699 (approved):
Note: we might be able to find only specific env into the specific container

Issue #207, opened 2020-11-30
Missing Context for Design Patterns


juhewitt:
**Describe the issue**
When registering a new project, the list of design patterns have no context as to what the design pattern is, though they can be assumed.

**Additional context**
As a user, I want to be able to find additional information on a design pattern, so as to know whether I'm currently using it, or it can reference me to an existing common component.

**Definition of done**
As a user, I click on a design pattern link, which launches a new window for the resource on the design pattern hosted on developer.gov.bc.ca.

![image.png](https://images.zenhubusercontent.com/5d658f800f25370001090f16/b921cef3-9005-4926-96cb-eaea9b601897)

juhewitt:
For the backlog, and future efforts.  This is based upon feedback from some community members that came to me with the suggestion.

w8896699:
Hi @juhewitt , is there any more specific requirement on this ticket, like which option need more explanation?

ShovelHand:
I don't understand; are there documentation resources for these currently on developer.gov.bc.ca, or is creating that part of the task?

ShovelHand:
With Theresa's help, I was able to find articles on most of these components from https://digital.gov.bc.ca/common-components.
I am still looking for documentation we have on Bambora, MS Teams notiications, file storage, and scheduling.

lukegonis:
Majority of components are completed, will merge what is complete to prod and complete the rest in a future sprint.

PR #659, opened 2022-03-10
#207 context for design patterns

ShovelHand:
Just added grey '?' elements that say "There is not yet any documentation linked for this component." when hovered over whenever There is not yet any documentation linked for [that] component.
image


okanji:
Hey @ShovelHand, this looks great! I like the greyed out question mark, it is intuitive. This looks good to me but it looks like you have some merge conflicts and checks to sort out above. Once those are done we can merge this in

okanji (changes requested):
Please fix the checks above

PR #661, opened 2022-03-14
Documentation link

ShovelHand:
image

From the task description:
"I could not find any links or references to any (Help or About) documentation that describes this applications purpose, functions, scope. Without this kind of information I find it harder to use the application and provide meaningful feedback." 

I just added a link to the top of the dashboard, "Git documentation for Platform Services Registry and a link to report issues can be found [here](http://localhost:8101/https://github.com/bcgov/platform-services-registry)"

I created a new React component "PSRDocumentationLink" for this, and added it on the dashboard.

### DIFF
--- a/web/src/components/dashboard/PSRDocumentationLink.tsx
+++ b/web/src/components/dashboard/PSRDocumentationLink.tsx
@@ -6,8 +6,8 @@ const PSRDocumentationLink: React.FC = () => {
   return (
     <Box>
       <Text>
-        Git documentation for Platform Services Registry and a link to report issues can be found
-        <Link to="https://github.com/bcgov/platform-services-registry">here</Link>
+        {/* eslint-disable-next-line */}
+        Git documentation for Platform Services Registry and a link to report issues can be found <Link to="https://github.com/bcgov/platform-services-registry">here</Link>
       </Text>
     </Box>
   );
--- a/web/src/components/dashboard/PSRDocumentationLink.tsx
+++ b/web/src/components/dashboard/PSRDocumentationLink.tsx
@@ -1,0 +1,16 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import { Box, Text } from 'rebass';
+
+const PSRDocumentationLink: React.FC = () => {
+  return (
+    <Box>
+      <Text>
+        Git documentation for Platform Services Registry and a link to report issues can be found
+        <Link to="https://github.com/bcgov/platform-services-registry">here</Link>
+      </Text>
+    </Box>
+  );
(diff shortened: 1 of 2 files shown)


okanji (approved):
Looks good @ShovelHand!

web/src/components/dashboard/PSRDocumentationLink.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import { Box, Text } from 'rebass';
+
+const PSRDocumentationLink: React.FC = () => {
+  return (
+    <Box>
+      <Text>
+        {/* eslint-disable-next-line */}
+        Git documentation for Platform Services Registry and a link to report issues can be found <Link to="https://github.com/bcgov/platform-services-registry">here</Link>
```
w8896699, line 10: I think we need a period at the end of the sentence. Also, why do we want to disable eslint for this line? Can we use variables for this URL? btw, I think it would be nice to create a new tab for this.

w8896699 (changes requested):
remmber run eslint before commit

okanji:
@w8896699  and @ShovelHand sorry I didn't see that Billy is also a reviewer otherwise I would have waited to approve. I agree with those comments you made, am not sure what you mean by new tab though.

What is the process when there are multiple reviewers on a PR? Do we wait to merge until all the reviewers have approved or do we only need one reviewer?

web/src/components/dashboard/PSRDocumentationLink.tsx
ShovelHand: I disabled eslint because it wanted me to either have the <Link> element on a new line, and threw an error if I had a whitespace after "found", thus I could not have the link appear as "found here", and we were stuck with "foundhere".

w8896699: There's a way to add white space in HTML format.

w8896699:
@okanji, sorry for the unclear comment, what I mean is we can set the target to blank. Because I learned that in bcgov website when we redirect user to a non-gov URL, we always want to open a new browser tab instead of changing current tab url.

web/src/components/dashboard/PSRDocumentationLink.tsx
ShovelHand: Oh, right. I forgot about that. I'll fix this.

w8896699: Please revert this pr and send a new pr with correct linting.

PR #662, opened 2022-03-14
Change invite to bc org logic

w8896699:
https://app.zenhub.com/workspaces/platform-experience-5bb7c5ab4b5806bc2beb9d15/issues/bcdevops/developer-experience/2364

I delete the api end point for invite githubUser to bcgov, move the logic to Backend after create request gets approved


okanji (approved):
Looks good to me, I always thought it would be better to do this in the back end so its nice that it has been moved, thank you

Issue #652, opened 2022-03-10
Show license plate in dashboard table


w8896699:
**Describe the issue**
Project License plate column is not currently included in the project list on the Dashboard page in the Project Registry. The user can manually add the license plate (aka namespace) column to the table using the button in the top right corner above the table. However, if the user goes to a project edit page and comes back to the Dashboard, the license plate column is removed from the table again.  

**Definition of done**
License plate column is permanently added to the project list on the Dashboard page as the last column.

ShovelHand:
Submitted for code review.

PR #666, opened 2022-03-17
License plate in dashboard #652

ShovelHand:
I'm embarrassed to say that this took me a long time because I didn't realize that the column was hidden by default. 
I replaced 'any' as parameter and return type accepted by fetchAllDashboardProjects. I will endeavour to do this for all code touched by my tickets' requirements. 
image


okanji (approved):
Thanks for making those type changes Alex

api/src/services/profile.ts
```diff
@@ -25,7 +27,9 @@ const dm = new DataManager(shared.pgPool);
 
 const { ProfileModel, ContactModel } = dm;
 
-const fetchAllDashboardProjects = async (userDetails: any): Promise<any> => {
+const fetchAllDashboardProjects = async (
+  userDetails: AuthenticatedUser
+): Promise<ProjectProfile> => {
```
w8896699, line 32: Nice I like it.

w8896699 (commented):
Looks good to me, I am wondering if we can remove package-lock change from this pr

okanji:
> Looks good to me, I am wondering if we can remove package-lock change from this pr

Ah yes, I don't think we're installing any new packages so we can exclude the package-lock

api/src/db/model/profile.ts
```diff
@@ -493,7 +493,7 @@ export default class ProfileModel extends Model {
 
     try {
       const results = await this.runQuery(query);
-      const { profiles } = results.pop();
+      const profiles = results.pop() as ProjectProfile;
```
w8896699, line 496: Changing the data structure will introduce bug

PR #677, opened 2022-04-06
Extend project detail and request tables

qrdn:
Removing the overflow modifier to each table component fixes the styling issues where the table is cut off on lower resolutions or minimized windows.
Screen Shot 2022-04-06 at 12 39 19 PM

### DIFF
--- a/web/src/components/dashboard/ProjectDetails.tsx
+++ b/web/src/components/dashboard/ProjectDetails.tsx
@@ -83,7 +83,7 @@ const ProjectDetails: React.FC<any> = (props) => {
 
   return (
     <>
-      <Box style={{ overflow: 'auto' }}>
+      <Box style={{}}>
         <Table
           columns={columns}
           data={data}
--- a/web/src/components/dashboard/ProjectRequests.tsx
+++ b/web/src/components/dashboard/ProjectRequests.tsx
@@ -161,7 +161,7 @@ const ProjectRequests: React.FC<any> = (props) => {
           />
         }
       />
-      <Box style={{ overflow: 'auto' }}>
+      <Box style={{}}>
         <Table
           columns={requestColumns}
           data={data}


web/src/components/dashboard/ProjectDetails.tsx
```diff
@@ -83,7 +83,7 @@ const ProjectDetails: React.FC<any> = (props) => {
 
   return (
     <>
-      <Box style={{ overflow: 'auto' }}>
+      <Box style={{}}>
```
w8896699, line 86: If you remove all style from it, then why we need to keep style props?

w8896699, line 86: also do you think this is the best solution? do we leave the right 0 margins while the left side has some margin(the table is not at the middle of the page.)

web/src/components/dashboard/ProjectRequests.tsx
```diff
@@ -161,7 +161,7 @@ const ProjectRequests: React.FC<any> = (props) => {
           />
         }
       />
-      <Box style={{ overflow: 'auto' }}>
```
w8896699, line 164: Does the request table also need to have this change? does it have the same problem?

w8896699 (changes requested):

web/src/components/dashboard/ProjectDetails.tsx
qrdn: I don't think it's possible to show the entire table in the window if it's minimized. The image may not have been very clear, but having no margin on the right side indicates that there is more to the table that the viewer can scroll over to.

qrdn: Right, I have made the changes

web/src/components/dashboard/ProjectRequests.tsx
qrdn: I changed the request table as well to have consistency within the two tables. If the request table were ever to gain more columns, then it will have the same behaviour as the project details table.

web/src/components/dashboard/ProjectDetails.tsx
w8896699: I have pulled this branch to my local and tested it. and made comment based on what I saw.

PR #672, opened 2022-03-25
Links to clusters and namespaces #508

ShovelHand:
Context: when looking at a projects details, the various namespaces it lives on (prod, dev, etc) will now be linked in the quota section.
image
While the change to users is slight, a few different areas of Platform Registry's infrastructure were changed to accomodate this, so attention is warranted.

PR #710, opened 2022-05-04
Update deploy.yaml

ShovelHand:
I don't see the parameters to add urls to namespaces when I look at a pod in OpenShift, so I added the parameters to deploy.yaml. I'm actually kind of guess that this was what was missing, so here it is for your review.


w8896699:
No, there's nothing todo with this deployment.yaml.  Change need to be made in https://github.com/bcgov-c/tenant-gitops-platform-registry/blob/c0762900a2dba9a259cf47dceef79baf2e4e97de/platform-registry-app/base/deployment.registry-api.yaml#L113.

w8896699:
And as I mentioned in RC, we only need one URL for each cluster(currently you have 4)

w8896699 (changes requested):
Please look at above two comments

ShovelHand:
So there's no doubt; you're idea is that we have two base  URLs, one for some Gold  namespace, and One for a silver namespace, then we just programatically add "-prod", or "-dev" or whatever to it as is appropriate, right @w8896699?

w8896699:
Yes

w8896699 (changes requested):
I suggest that you revert the previous commit and redo the whole task in a new pr as the whole logic now has been changed. https://github.com/bcgov/platform-services-registry/pull/672

api/src/controllers/quota.ts
```diff
@@ -77,24 +77,15 @@ export const getAllAllowedQuotaSize = async (req: Request, res: Response) => {
...
-    const prodGoldUrl = process.env.PRODUCTION_NAMESPACE_GOLD_URL || "";
-    const devGoldUrl = process.env.DEVELOPMENT_NAMESPACE_GOLD_URL || "";
-    const testGoldUrl = process.env.TEST_NAMESPACE_GOLD_URL || "";
-    const toolsGoldUrl = process.env.TOOLS_NAMESPACE_GOLD_URL || "";
-    // silver
-    const prodSilverUrl = process.env.PRODUCTION_NAMESPACE_SILVER_URL || "";
-    const devSilverUrl = process.env.DEVELOPMENT_NAMESPACE_SILVER_URL || "";
-    const testSilverUrl = process.env.TEST_NAMESPACE_SILVER_URL || "";
-    const toolsSilverUrl = process.env.TOOLS_NAMESPACE_SILVER_URL || "";
+    const goldUrl = process.env.GOLD_CLUSTER_URL || "";
```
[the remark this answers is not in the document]
ShovelHand: Yes

ShovelHand:
I am not redoing this task as a new PR. That is a needless reproduction of effort on something that has already gone on for way too long.

web/src/components/profileEdit/QuotaCard.tsx
```diff
@@ -26,7 +26,6 @@ import {
   ProjectNamespaceResourceQuotaSize,
 } from '../../types';
 import { BaseIcon } from '../common/UI/Icon';
-import useRegistryApi from '../../hooks/useRegistryApi';
```
w8896699, line 29: clean up function in useRegistryApi as well

w8896699: and in web/src/views/ProfileEdit.tsx

api/src/router/routes/quota.ts
```diff
@@ -31,7 +31,7 @@ const router = express.Router();
 // Fetches raw quota specifications from DB
 // This endpoint is not being utilized by the front end; implemented for future use
 router.get("/", asyncMiddleware(fetchQuota));
-router.get("/namespaceUrls", asyncMiddleware(fetchNamespaceUrls));
+router.get("/namespaceUrls", asyncMiddleware(getClusterUrls));
```
w8896699, line 34: Are we still need changes in api?

w8896699 (changes requested):
Almost there

ShovelHand:
I took a good, close look, and I think we're all good now. I thought I'd get you to have a last look before merging @w8896699

w8896699 (approved):

PR #726, opened 2022-06-01
changed project to product in most components/views

qrdn:
This changes instances of "project" to "product" on the Registry Frontend

I also updated the snapshots for the jest tests, which I'm not sure is the correct procedure.


w8896699 (approved):
LGTM

PR #755, opened 2022-07-27
added dashes and spaces availability for names

qrdn:
This fixes the issue with the inclusion of spaces and dashes in names.


w8896699 (approved):
LGTM

PR #767, opened 2022-08-30
quata change checkbox ignore

### LINKED ISSUE #728: Quota update checkbox bug
**Describe the issue**
On the Resource Quotas form, clicking the confirmation checkbox will dirty the form, allowing the user to submit a request for different resource quotas without any change to the quota form fields (CPU, Memory, Storage, or Snapshot).

**Additional context**
After opening the Resource Quotas form (which holds [the `<QuotaCardEdit>` component](https://github.com/bcgov/platform-services-registry/blob/master/web/src/components/profileEdit/QuotaCardEdit.tsx)), before making any changes to the form state:
QuotaCardEdit component with no changes

After selecting the checkbox generated by [this `<Field>` component](https://github.com/bcgov/platform-services-registry/blob/master/web/src/components/profileEdit/QuotaCardEdit.tsx#L499), the form is now able to be submitted even though no changes to the quota fields have been made:
QuotaCardEdit component after selecting the checkbox but without changing any of the quota fields in the form

De-selecting the checkbox will generate a "Required" field validation error as we would like, but the form submission button is still enabled:

Look at [the `<EditSubmitButton>` being instantiated here](https://github.com/bcgov/platform-services-registry/blob/master/web/src/components/profileEdit/QuotaCardEdit.tsx#L508) and track its props from [the `<ProfileEdit>` view](https://github.com/bcgov/platform-services-registry/blob/master/web/src/views/ProfileEdit.tsx).

**Definition of done**
- [x] A user should have to change at least one other form field in addition to selecting the checkbox before the submit button is enabled.
- [x] The submit button should not be enabled if the user changes a form field to a new value but then changes it back to its original value. The logic for enabling the submit button should be comparing the original form state to the current state.
- [ ] Integrate into ArgoCD

Kolezhanchik:
Closes #728  added additional check on disabled button based on react final form properties

### DIFF
--- a/web/src/components/common/UI/EditSubmitButton.tsx
+++ b/web/src/components/common/UI/EditSubmitButton.tsx
@@ -21,7 +21,7 @@ import { StyledFormButton, StyledFormDisabledButton } from './Button';
 export const EditSubmitButton = (props: any) => {
   const { hasPendingEdit, isProvisioned, pristine, touched } = props;
 
-if (pristine|| Object.values(touched).filter(Boolean).length===1) {
+  if (pristine || Object.values(touched).filter(Boolean).length === 1) {
     return (
       <>
         <StyledFormDisabledButton disabled style={{ display: 'block' }}>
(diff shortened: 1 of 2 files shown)


Kolezhanchik:
Hi @w8896699 can you please, review the changes? thanks

w8896699 (approved):
Looks good to me

PR #769, opened 2022-09-06
fixed problems encountered when building api node packages

ShovelHand:
These issues are likely to do with versions of node, etc. I'm using, and should be vetted thoroughly before being added to main.
Versions:
Node v12.22.12
npm 6.14.16
express 4.18.1

### DIFF
--- a/api/src/controllers/profile.ts
+++ b/api/src/controllers/profile.ts
@@ -94,7 +94,7 @@ export const updateProfileContacts = async (
     user,
   }: { params: any; body: Contact[]; user: AuthenticatedUser },
   res: Response
-): Promise<void> => {
+): Promise<Response> => {
   const { ContactModel, RequestModel } = dm;
   const { profileId } = params;
   const contacts = body;
--- a/api/src/server.ts
+++ b/api/src/server.ts
@@ -23,16 +23,17 @@ import app from "./index";
 const env = config.get("environment");
 const port = config.get("port");
 
-app.listen(port, "0.0.0.0", (err) => {
-  if (err) {
+app.listen(port, "0.0.0.0", 0, () => {
+  try {
+    if (env !== "production") {
+      return started(port);
+    }
+    return logger.info(`Production server running on port: ${port}`);
+  } catch(err) {
     return logger.error(
       `There was a problem starting the server, ${err.message}`
     );
   }
-  if (env !== "production") {
-    return started(port);
-  }
-  return logger.info(`Production server running on port: ${port}`);
 });
 
 module.exports = app;
--- a/api/src/controllers/communications/helpers.ts
+++ b/api/src/controllers/communications/helpers.ts
@@ -14,7 +14,7 @@ export const getToken = async () => {
 
   const params = {
     client_id: "mautic-subscription-api",
-    client_secret: keycloakClientSecret,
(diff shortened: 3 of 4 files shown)


api/src/controllers/profile.ts
```diff
@@ -94,7 +94,7 @@ export const updateProfileContacts = async (
     user,
   }: { params: any; body: Contact[]; user: AuthenticatedUser },
   res: Response
-): Promise<void> => {
+): Promise<Response> => {
```
ty2k, line 97: I like the typing idea you have with this change (and the equivalent change in `request.ts`), but it looks like it's [breaking the build](https://github.com/bcgov/platform-services-registry/runs/8215361327?check_suite_focus=true).

With this change, the TypeScript compiler is giving me this error: `Type 'void' is not assignable to type 'Response<any>'.` It looks like this is due to the `Response` method we are calling: `.end()`. I actually had to look that one up - it's not something I've ever used in all my years of using Express. See [`res.end()`](https://expressjs.com/en/api.html#res.end) vs [`res.send()`](http://expressjs.com/en/4x/api.html#res.send).

I'm not enough of a TypeScript wizard to know what's the best way to go about solving this, but I see this [StackOverflow answer](https://stackoverflow.com/a/56080762) points to a reason for why this happens with `.end()`.

I notice if I naively change all the `.end()` calls to `.send()`, the TypeScript compiler is happier and I can run a build successfully. However, it's not safe to just make this change, as it causes new test breakages in `profile.spec.ts`.

api/src/server.ts
```diff
@@ -23,16 +23,17 @@ import app from "./index";
 const env = config.get("environment");
 const port = config.get("port");
 
-app.listen(port, "0.0.0.0", (err) => {
-  if (err) {
+app.listen(port, "0.0.0.0", 0, () => {
```
ty2k, line 26: The new `try`/`catch` block looks good, but the new `0` argument for `backlog` should be left out. See the [Node.js `server.listen()` documentation](https://nodejs.org/api/net.html#serverlisten) for details. A value of `0` effectively limits us to no queue at all, whereas we just want to take whatever the default is from the environment.

ty2k (changes requested):
Thanks for trying to improve this! 💪 

Beyond the changes I've listed in the inline comments, it would be prudent to make sure we are using the same Node version across the `./api` directory if you are confident that is giving you issues.

Here are some different places with different values:
- [`package.json`](https://github.com/bcgov/platform-services-registry/blob/master/api/package.json#L86) has 10 in the `engines` field
- [`Dockerfile`](https://github.com/bcgov/platform-services-registry/blob/master/api/Dockerfile#L1) lists 10.14.x

<s>I don't see a reference to Node 12 in this project. I do notice that `./db/Dockerfile` is using `postgres:12.2-alpine`, which doesn't refer to Node, but might have been the original source of thinking it was necessary to be on 12.x?</s>

Scratch that, I actually see references to Node at 10 and 12 in the workflows in the `./.github` directory. It would be a good idea to use 12 everywhere if that is what is being used to build these images. Might need more research on this.

You also mention using a newer version of Express, but there's no update to `package.json` and `package-lock.json` in this PR. Please add those if that's the intent.

PR #782, opened 2022-09-27
Pass touched prop to all EditSubmitButton instances

ty2k:
Currently, editing a text field in `<ProfileCardEdit>` or `<ContactCardEdit>` will cause a user to see a blank screen, and they will need to refresh the page to fix it. This looks to be caused by a missing `touched` prop in the `<EditSubmitButton>` component which was updated in #767 . This is the console error that will appear:

```
Uncaught TypeError: Cannot convert undefined or null to object
    at Function.values (<anonymous>)
    at Sa (EditSubmitButton.tsx:24:26)
```

This pull request passes the `touched` prop from `react-final-form` to all instances of the `<EditSubmitButton>` to fix the issue.

### DIFF
--- a/web/src/components/common/UI/EditSubmitButton.tsx
+++ b/web/src/components/common/UI/EditSubmitButton.tsx
@@ -19,9 +19,9 @@ import React from 'react';
 import { StyledFormButton, StyledFormDisabledButton } from './Button';
 
 export const EditSubmitButton = (props: any) => {
-  const { hasPendingEdit, isProvisioned, pristine, touched } = props;
+  const { hasPendingEdit, isProvisioned, pristine } = props;
 
-  if (pristine || Object.values(touched).filter(Boolean).length === 1) {
+  if (pristine) {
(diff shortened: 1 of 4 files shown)


ty2k:
Converting this to draft, as I have found a case where I can't submit a change after making a valid edit (to my display name in the `<ContactCardEdit>` component. I wonder if we should revert the changes from #767 and re-introduce the quota form bug instead?

okanji:
@ty2k let me look at that quota form bug issue and I'll get back to you

okanji:
@ty2k, to me, this bug that 767 introduces is worse worse than what it tries to fix so I agree that we are better off reverting it

ty2k:
@okanji I will push new commits to do that here ASAP.

ty2k:
I'm closing this without merging as @okanji and I decided it would be better to revert the problematic pull request rather than introduce new commits to do effectively the same thing in this PR.

PR #783, opened 2022-09-27
Revert "quata change checkbox ignore"

okanji:
Reverts bcgov/platform-services-registry#767

This PR introduced a bug described in #782


ty2k:
This is a better solution than to add new commits like in #782 . While this re-introduces the quota checkbox form bug, it will at least stop the front-end from crashing.

ty2k (approved):

PR #786, opened 2022-10-11
Idir instead of GitHub

ShovelHand:
This change involves getting IDIR info for contacts on project requests instead of GitHub.
image
 
The best way to test is to start writing in the last name of a IDIR user in the appropriate box, and we should get back their user info and a user Photo if they have on. Looks like most users don't have one. Try typing in 'mitovska' to see our wonderful boss' face an example of this working properly. 

There are a few user experience issues I've created separate tickets to address. These are:
- having a better placeholder for missing user photos instead of a broken image link and alt text. 
- Let's get rid of the "Idir Id" field altogether, and just use the email address to get the user. I haven't done this yet because it involves a bit of back-end work, and it's time to move this task along.


ShovelHand:
@ty2k, @okanji Would you guys mind having a look? Tyler is most familiar with the changes made since he gave me so much help, but it's a big PR, and another pair of eyes is a good idea.

ty2k:
I haven't been able to test this in the lab today due what looks like some kind of infinite loop of network requests in the new product creation flow. I'm not sure if this is a connection issue on my end, or some circular logic in the React component (ex: a network request in the component is setting state in a way that is ultimately changing the component's own props, causing it to render over and over). I think this might be what you were talking to Oamar about, but I'm not 100% sure:

Screen Shot 2022-11-17 at 4 22 09 PM

I notice that I don't experience this issue in the product edit flow. However, it looks like we need to update this view to account for the new IDIR requirements (dropping references to GitHub if users are unable to select GitHub users, etc).

Screen Shot 2022-11-17 at 3 50 29 PM

web/src/redux/githubID/graphAuthConfig.js
```diff
@@ -0,0 +1,21 @@
+/* this is used by the (hopefully silent) OAuth flow to get a token to access Microsoft Graph's API assigning contacts 
+to project requests. It should not require editing. You can find the Azure project, where the Tenant and Client IDs are
+defined, here: https://portal.azure.com/#view/Microsoft_AAD_RegisteredApps/ApplicationMenuBlade/~/Overview/appId/5afdfc62-637b-41cf-b186-b2de816faaf9/isMSAApp~/false
+or look for 'Platform Services Registry' on BC Gov's Azure environment. */
+
+export const msalConfig = {
+  auth: {
+    clientId: '5afdfc62-637b-41cf-b186-b2de816faaf9',
```
ty2k, line 8: Let's access these values from environment variables set in `.env` instead. Since this React application was scaffolded with Create React App, we need to follow their way of doing this: https://create-react-app.dev/docs/adding-custom-environment-variables/

Your environment variables in `.env` should look like `REACT_APP_AZURE_CLIENT_ID` and `REACT_APP_AZURE_TENANT_ID`. The `REACT_APP_` prefix is necessary for any environment variables to show up in an app created with Create React App.

.env.example
```diff
@@ -26,3 +26,9 @@ CHES_BASEURL=
 CHES_SSO_TOKEN_URL=
 CHES_SSO_CLIENT_ID=
 CHES_SSO_CLIENT_SECRET=
+
+# Microsoft Azure Application Details
+AZURE_CLIENT_ID=
+#May be called "password" rather than "secret" in the authoriaztion details
+AZURE_CLIENT_SECRET=
```
ty2k, line 33: I'm see this addition to the `.env.example`, but I don't see these variables being used, except when they're passed through the `docker-compose` file. I think it's a good idea to leave these here and use them in the React front-end.

ty2k:
Can you please update the placeholder text in `<GitHubUserValidation>` to indicate that we need an IDIR email address as opposed to a GitHub username?

Screenshot 2022-11-23 at 2 39 39 PM

ty2k:
I've added an example branch [graph-api-response-logic](https://github.com/bcgov/platform-services-registry/commits/graph-api-response-logic) with two commits where I updated `searchIdirUsers()` in `githubID.action.ts` and the `GITHUBID_USER_DOES_NOT_EXIST` case in `githubID.reducer.ts` to bring back the "No user found :(" message when a search fails, as opposed to the current state with leaves the "Loading..." message there.

I'd recommend doing something like this to let users know that the search was completed but no results were found. If we were to deploy with the "Loading..." message staying up after a search finishes, it will cause confusion.

ProfileCreate when searching for a user that doesn't exist

ty2k:
There are at least three leftover labels referencing GitHub usernames that we need to update:

- [ ] Update the "Write a github username" placeholder text in `GithubUserValidation.tsx`
- [ ] Replace references to "GitHub Id" in `ContactCardEdit.tsx`
- [ ] Replace reference to "Github IDs" in `PublicLanding.tsx`

web/package.json
```diff
@@ -27,6 +30,7 @@
     "keycloak-js": "^9.0.3",
     "prop-types": "^15.7.2",
     "react": "^16.13.1",
+    "react-bootstrap": "^2.5.0",
```
ty2k, line 33: We're importing the minified distribution code from the `bootstrap` package in `web/src/index.tsx`, but we don't actually import `react-bootstrap` anywhere. `react-bootstrap` should be left out unless we're using the components from that library directly, I think.

web/src/views/ProfileCreate.tsx
```diff
@@ -113,4 +161,19 @@ const ProfileCreate: React.FC = () => {
   );
 };
 
-export default ProfileCreate;
+const mapStateToProps = (state: any, githubID: any) => ({
+  technicalLead1: selectTechnicalLead(0)(state),
+  technicalLead2: selectTechnicalLead(1)(state),
+  stateProductOwner: selectProductOwner()(state),
+});
+// const mapDispatchToProps = (dispatch: any) => ({
```
ty2k, line 169: I'm assuming you won't be using this commented code for anything, right? It should be left out unless it's necessary.

ty2k (changes requested):
Thanks for the hard work on this major functional change, @ShovelHand!

I have tested the Registry stack locally in its current state using these manual test cases:

1. Try to submit a new request with known good IDIRs for the PO, TL1, and TL2. (Should succeed)
2. Try to submit a new request with known good IDIRs for the PO and TL1. (Should succeed)
3. Try to submit a new request with known good IDIRs for the PO and TL1, but not for TL2. (Should not be possible)
4. Try to submit a new request with known good IDIR for the PO, but not for TL1. (Should not be possible)
5. Try to submit an edit request with known good IDIRs for the PO, TL1, and TL2. (Should succeed)
6. Try to submit an edit request with known good IDIRs for the PO and TL1. (Should succeed)
7. Try to submit an edit request with known good IDIR for the PO and TL1, but not for TL2. (Should not be possible)
8. Try to submit an edit request with known good IDIR for the PO, but not for TL1. (Should not be possible)

I find that all of these manual test cases work with the code at this point.

That being said, we should address the issues I listed in the thread above before deploying. We should also finish [#3099 Document Rollout Strategy for replacing Github IDs for IDIR for POs and TLs in Registry](https://app.zenhub.com/workspaces/platform-experience-5bb7c5ab4b5806bc2beb9d15/issues/bcdevops/developer-experience/3099) ahead of deploying this, I assume?

It might also be good to deploy [#3147 Gracefully handle missing photos when Idir user added to Project Request in Platform Services Registry](https://app.zenhub.com/workspaces/platform-experience-5bb7c5ab4b5806bc2beb9d15/issues/bcdevops/developer-experience/3147) before or at the same time as this change, since the large majority of IDIR users aren't going to have avatars (vs GitHub where every user gets an auto-generated pixel art avatar). This should be a straightforward change compared to the much more involved work in this PR. 😁

web/src/__tests__/ProfileCreate.test.tsx
```diff
@@ -14,14 +14,14 @@
...
-import { Provider } from 'react-redux';
-import { store } from '../redux/store';
-import ProfileCreate from '../views/ProfileCreate';
+// import React from 'react';
+// import { Provider } from 'react-redux';
 import { MsalReactTester } from 'msal-react-tester';
-import { MsalProvider } from '@azure/msal-react';
+// import { MsalProvider } from '@azure/msal-react';
+// import { store } from '../redux/store';
+// import ProfileCreate from '../views/ProfileCreate';
```
w8896699, line 24: Please try not to merge in comment out code, and also it seems like you disable this test, you can probably disable/skip this test instead. 
Last, please squash before merge to master branch

w8896699 (changes requested):

PR #819, opened 2022-12-14
div with classname appears when profiles fetched

ShovelHand:
Hi, Billy. 
I think this is what you had in mind, but one thing I'm not sure about is how searching for elements by className in React works for external stuff, like entity. 
Let me know if you need me to rework this.

### DIFF
--- a/web/src/views/Dashboard.tsx
+++ b/web/src/views/Dashboard.tsx
@@ -16,7 +16,6 @@
 
 import { useKeycloak } from '@react-keycloak/web';
 import React, { useEffect, useState } from 'react';
-import { Bool } from 'reselect/es/types';
 import ProjectDetails from '../components/dashboard/ProjectDetails';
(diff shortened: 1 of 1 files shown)


ShovelHand:
duplicate of 794

web/src/views/Dashboard.tsx
```diff
@@ -16,6 +16,7 @@
 
 import { useKeycloak } from '@react-keycloak/web';
 import React, { useEffect, useState } from 'react';
+import { Bool } from 'reselect/es/types';
```
w8896699, line 19: You are not using this Bool why add here?

w8896699 (changes requested):
Almost there

web/src/views/Dashboard.tsx
ShovelHand: Woops. Removed that.

PR #1032, opened 2023-09-19
1025 login UI

calbertan:
The navbar redirect + login UI

should be ok


funtigr:
@Kolezhanchik 
Could you please provide a link with actual wording for the Login Page.
An could you please look at the implementation regarding Figma Mockups.

okanji (approved):

okanji:
@funtigr , I've approved this so that Christopher can show it in the sprint review. We can change the wording later in a new ticket. Thanks :)

PR #1108, opened 2023-10-17
chore: add pre-commit config

junminahn:


okanji:
Great! It will be helpful to have these checks for our repo

PR #1119, opened 2023-10-19
chore: fix build errors

junminahn:
I believe it could happen with a `prisma`'s nature of auto-generating types by running `prisma generate`.
So that when `schema.prisma` is updated and forgets to run the command again, it would be difficult to pick up the type errors in the local environments.

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -148,6 +148,7 @@ model PublicCloudRequestedProject {
   created                  DateTime            @default(now())
   accountCoding            String
   budget                   Budget
+  billingGroup             String
   projectOwnerId           String              @db.ObjectId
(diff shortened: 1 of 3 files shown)


prisma/schema.prisma
```diff
@@ -148,6 +148,7 @@ model PublicCloudRequestedProject {
   created                  DateTime            @default(now())
   accountCoding            String
   budget                   Budget
+  billingGroup             String
```
okanji, line 151: I don't believe that we have a billing group

okanji: Actually, we changed billing group to account coding, so we would need to remove billing group from the seed file and replace it with account coding

okanji (approved):

PR #1227, opened 2023-10-31
public cloud create modal added

Kolezhanchik:


components/form/AccountCoding.tsx
```diff
@@ -18,19 +18,19 @@ export default function AccountCoding({
 
   const [accountCoding, setAccountCoding] = useState({
```
junminahn, line 19: I see the same logic below to parse `accountCodingInitial` to produce `accountCoding` object;
I think we can make it as a function to reuse in multiple places.

junminahn (commented):
Looking good to me and thank you for running `pre-commit hooks` to reduce the `commit noises`. I just left a comment on it.

PR #1236, opened 2023-11-01
feat: send emails for create/approve/reject and more tests

calbertan:


app/api/private-cloud/decision/[licencePlate]/route.ts
```diff
@@ -100,8 +103,92 @@ export async function POST(req: NextRequest, { params }: { params: Params }) {
   await subscribeUsersToMautic(users, request.requestedProject.cluster, 'Private');
 
   // Send emails
+  sendRequestApprovalEmails(request);
 
   return new NextResponse(`Decision request for ${request.licencePlate} succesfully created.`, {
     status: 200,
   });
 }
+function sendRequestRejectionEmails(
```
Comment (junminahn) [inline app/api/private-cloud/decision/[licencePlate]/route.ts:112]: Can we have distinct types for improved reusability and readability?

ches/emailHandler.ts
```diff
@@ -15,25 +18,59 @@ export const sendCreateRequestEmails = async (formData: PrivateCloudCreateReques
...
+    await chesService.send({
+      bodyType: 'html',
+      body: email,
+      // For all project contacts. Sent when the project set deletion request is successfully submitted
+      to: [
+        request.requestedProject.projectOwner,
+        request.requestedProject.primaryTechnicalLead,
+        request.requestedProject.secondaryTechnicalLead,
+      ]
+        .filter(Boolean)
```
junminahn, line 47: Could we relocate this `filtering` process to the `ches service level`, as it seems to be duplicated in multiple places?

ches/emailHandler.ts
```diff
@@ -15,25 +18,59 @@ export const sendCreateRequestEmails = async (formData: PrivateCloudCreateReques
...
+  } catch (error) {
+    console.error(error);
+  }
+};
+
+export const sendRequestApprovalEmails = async (request: PrivateCloudRequestWithRequestedProject) => {
+  const email = render(RequestApproval({ request }), { pretty: true });
+  try {
+    await chesService.send({
+      bodyType: 'html',
```
junminahn, line 39: I understand that the majority of templates are in `html` format. Would it be possible to set html as the default `bodyType` value in the `ches service`, allowing for the option to override it when calling the send function?

ches/emailHandler.ts
```diff
@@ -15,25 +18,59 @@ export const sendCreateRequestEmails = async (formData: PrivateCloudCreateReques
...
+      body: email,
+      // For all project contacts. Sent when the project set deletion request is successfully submitted
+      to: [
+        request.requestedProject.projectOwner,
+        request.requestedProject.primaryTechnicalLead,
+        request.requestedProject.secondaryTechnicalLead,
+      ]
+        .filter(Boolean)
```
junminahn, line 69: the same comment as above.

emails/NewRequest.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { TailwindConfig } from './components/TailwindConfig';
+const defaultTheme = require('tailwindcss/defaultTheme');
+
+interface EmailProp {
+  formData: PrivateCloudCreateRequestBody;
+}
+
+export const NewRequest = ({ formData }: EmailProp) => {
+  if (!formData) {
+    formData = sampleFormData;
```
junminahn, line 19: It seems that the fallback to `sampleFormData` is intended to support `local dev or dev` environments. However, I believe it would be better to throw an error or return an empty string if no data is passed into the template.

emails/RequestApproval.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { sampleRequest } from './components/Params';
+import { TailwindConfig } from './components/TailwindConfig';
+
+interface EmailProp {
+  request: PrivateCloudRequestWithRequestedProject;
+}
+
+export const RequestApproval = ({ request }: EmailProp) => {
+  if (!request) {
+    request = sampleRequest;
```
junminahn, line 18: the same comment as above

emails/RequestRejection.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { sampleRequest } from './components/Params';
+import { TailwindConfig } from './components/TailwindConfig';
+
+interface EmailProp {
+  request: PrivateCloudRequestWithRequestedProject;
+}
+
+export const RequestRejection = ({ request }: EmailProp) => {
+  if (!request) {
+    request = sampleRequest;
```
junminahn, line 18: the same comment as above

emails/components/Header.tsx
```diff
@@ -3,12 +3,13 @@ import { Img, Text } from '@react-email/components';
 export default function Header() {
   return (
     <div className="flex flex-row border-b-3 border-bcorange bg-bcblue shadow">
+      {/* <Img src="/static/logo.png" alt="BC Platform Services Product Registry" width={58} height={41} className="m-auto ml-2 mr-2" /> */}
       <Img
-        src="/static/logo.png"
+        src="https://dev-pltsvc.apps.silver.devops.gov.bc.ca/private-cloud/products/logo.png"
```
junminahn, line 8: I have a concern about supporting other environments like `test or production`.
Introducing another environment variable, perhaps named `APP_URL`, could be a good solution. This variable could be used in the template, allowing the source to be defined as `${process.env.APP_URL}/private-cloud/products/logo.png`.

emails/components/NamespaceDetails.tsx
```diff
@@ -0,0 +1,28 @@
+import Image from 'next/image';
+import Logo from '../assets/logo.png';
+import { Button, Heading, Img, Link, Text } from '@react-email/components';
+import { PrivateCloudCreateRequestBody } from '@/schema';
+
+export default function NamespaceDetails({ cluster, licencePlate }: { cluster: any; licencePlate?: any }) {
```
junminahn, line 6: Let's try to avoid using `any` type wherever possible. I believe both arguments are of type `string`, correct?

emails/components/ProductDetails.tsx
```diff
@@ -3,42 +3,44 @@ import Logo from '../assets/logo.png';
...
-export default function Header({ data }: EmailProp) {
+export default function ProductDetails({
+  name,
+  description,
+  ministry,
+  po,
+  tl1,
+  tl2,
+}: {
+  name: String;
```
junminahn, line 14: I'm not certain if `String` with a capital 'S' is a valid TypeScript type. Additionally, let's try to minimize the usage of the `any` type wherever possible in the code below.

emails/components/TailwindConfig.ts
```diff
@@ -0,0 +1,40 @@
...
+      colors: {
+        bcblue: '#003366',
+        bcorange: '#FCBA19',
+        darkergrey: '#344054',
+        mediumgrey: '#475467',
+        tableheadergrey: 'rgba(214, 216, 213, 0.15)',
+        tablebordergrey: '#EAECF0',
+        disabledborder: 'D0D5DD',
+        cloudgrey: '#667085',
+        divider: '#0000001f',
```
junminahn, line 20: Whatever the case may be, let's keep consistent case conventions. I've noticed that all the color hash strings elsewhere are in uppercase.

junminahn (commented):
Great job overall! 👍 I've added a few comments on the PR. Utilizing snapshot testing is a smart choice, as it's much easier to maintain.

I believe our peer review process could be even more efficient if we aim for more frequent, smaller PRs as we progress. This approach would reduce the likelihood of encountering code conflicts and help address any potential misdirection in the early stages of the work.

ches/emailHandler.ts
calbertan: Since we are creating each function for specific requests/emails, I don't think that will be necessary since we will create a separate function anyways if we need to send a different bodyType.

junminahn: What I'm suggesting is to establish a default value for the `bodyType` option within the send function. Default values are used when specific options are not provided, making them useful in most cases.
I see `email: any` in `send` function; it would be nicer to define the email type there as well.

calbertan:
removed the snapshot tests for various reasons:
1. requires you to re-snapshot on any single UI change 
2. changes/warnings are not the easiest to understand
3. updating snapshots is necessary before any push with UI changes

This leads to blindly updating snapshots whenever a UI change is made because updating snapshots are something you have to do anyways. I don't think this achieves anything. 

Maybe this will be more useful once everything is completed, when not so many changes are needed to be made at a time

junminahn (commented):
👍

junminahn (approved):

PR #1239, opened 2023-11-02
fix(1235): resolve node-fetch Headers importing issue

junminahn:


okanji (commented):
Hi Junmin, I have just realized something. This code was copy-pasted from our previous node.js app by me. Now that we are using next.js, I think `fetch` is provided out of the box and we might not need to import `node-fetch`. Maybe we can try simply using `fetch`?

junminahn:
> Hi @junminahn , I have just realized something. This code was copy-pasted from our previous node.js app by me. Now that we are using next.js, I think `fetch` is provided out of the box and we might not need to import `node-fetch`. Maybe we can try simply using `fetch`?

@okanji as far as I know, `fetch` is not available in NodeJS, so we need to use either `node-fetch` or `axios`.

okanji:
> > Hi @junminahn , I have just realized something. This code was copy-pasted from our previous node.js app by me. Now that we are using next.js, I think `fetch` is provided out of the box and we might not need to import `node-fetch`. Maybe we can try simply using `fetch`?
> 
> @okanji as far as I know, `fetch` is not available in NodeJS, so we need to use either `node-fetch` or `axios`.

Ah yes! Thanks

okanji (approved):

PR #1247, opened 2023-11-03
refactor(1235): update useQuery syntax accordingly to the new version

junminahn:
- ref: https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5#supports-a-single-signature-one-object


okanji (approved):
Great. It's nice that we are using the latest version and syntax. We will need to implement some common error handling logic in the future.

PR #1263, opened 2023-11-06
tests for queries

Amritpal-Nijjar:
created tests for the privateCloudProjects function to ensure proper filtering


junminahn (commented):
It seems that our primary focus for this particular PR is on generating queries for CSV data. I appreciate that I can easily follow your changes in the code.

As a standard best practice, let's make sure to stick to the `DRY principle` for improved code reusability, maintainability, and to minimize redundancy. Here are my recommendations:

1. Move the `paginatedQueries` into the `queries` directory, since paginated queries are also a part of the high-level `queries`.
2. Organize all `types` used in the query files to `eliminate duplicates`.
3. Utilize the `paginated query` as a foundational function for `privateCloudProjects`, allowing it to accept optional `pageSize` and `pageNumber` parameters. This way, `privateCloudProjectsPaginated` can use the base function with extended options.

Amritpal-Nijjar:
I get an error when moving paginatedQueries into the queries directory.

junminahn:
> I get an error when moving paginatedQueries into the queries directory.

I believe there are a few more places need to be updated accordingly. Are you able to debug that?

Amritpal-Nijjar:
> > I get an error when moving paginatedQueries into the queries directory.
> 
> I believe there are a few more places need to be updated accordingly. Are you able to debug that?
Unfortunately, no success...

junminahn:
> > > I get an error when moving paginatedQueries into the queries directory.
> > 
> > 
> > I believe there are a few more places need to be updated accordingly. Are you able to debug that?
> > Unfortunately, no success...

I see, we can take a look together; revert what you have been doing and try making changes on other comments; I will let you know when I am available for a chat.

junminahn (approved):

PR #1305, opened 2023-11-09
Updated comments on the test cases, removed duplicate test

Amritpal-Nijjar:
Added more descriptive comments to all test cases, as well as removed a duplicate test case. Started setup for csv API route.


junminahn (commented):
I have a couple of general comments:

1. It's great that we're checking for certain parameters' existence in the routes.
However, I suggest adding more extensive validation to the route, such as verifying if the cluster is one of the allowed ones or if the email address is valid. You can refer to this blog post for reference: [[link](https://www.kirandev.com/nextjs-api-routes-zod-validation)](https://www.kirandev.com/nextjs-api-routes-zod-validation).

3. In general, additional comments may not be necessary as long as the codebase itself is self-explanatory. This is especially true in test scenarios, where there is a description section to provide enough context about the test behavior. For example, the following block:
```
  // Verifies that all projects can be retrieved without any filter.
  test('Should return all projects ', async () => {
    const projects = await privateCloudProjects();
    expect(projects.length).toBe(5);
  });
```
can be updated without a comment like this:
```
  test('shoud return all projects without any filter ', async () => {
    const projects = await privateCloudProjects();
    expect(projects.length).toBe(5);
  });
```

Amritpal-Nijjar:
image

Amritpal-Nijjar:
Screenshot 2023-11-16 at 1 48 03 AM

junminahn (approved):

PR #1342, opened 2023-11-15
feat(897): add rows per page selection on paginations

### LINKED ISSUE #897: [New UI] Rows per page feature adjustments
**Description**
Since we are moving to a new design the task description changes.

We need to implement the '**Rows per page**' feature.
The case previously described by Matt must be taken into account:
**Given** User changes the default '**Rows per page**' to a non-default value (e.g. 50)
**And** User goes to the second (or any non-first) page
**When** User clicks into a Product
**And** returns back to the Products List page
**Then** the user should be returned to the same page with the same set of '**Rows per page**'

**Definition of done**
If this feature suggestion is adopted: 

- [ ] The number of rows selected is retained when switching pages

junminahn:
closes #897 
image


okanji:
💯

okanji (approved):
Awesome. Thanks for attaching a screen shot for the ui, looks good!

PR #1361, opened 2023-11-17
feat: add ministry role access logic

junminahn:


queries/user.ts
```diff
@@ -1,5 +1,25 @@
...
 import prisma from '@/lib/prisma';
+import { getServerSession } from 'next-auth/next';
+import { authOptions } from '@/app/api/auth/options';
+import checkUserMinistryRole from '@/components/utils/checkUserMinistryRole';
+
+export async function userInfo() {
+  const session = await getServerSession(authOptions);
+  const { email: authEmail, roles: authRoles } = session
+    ? session.user
+    : {
```
junminahn, line 11: I believe we should return `null` if session does not exist and the `query` should return `null` as well if no session.
Because when `email` is undefined, it will consider the user as the admin user in the query generation.

queries/private-cloud/helpers.ts
```diff
@@ -182,7 +204,7 @@ export async function getPrivateCloudProjectsResult({
           as: 'secondaryTechnicalLeadDetails',
         },
       },
-      { $match: searchQuery },
+      { $match: await searchQuery },
```
junminahn, line 207: Let's takes `query object` rather than `Promise[query object]`.

components/utils/checkUserMinistryRole.tsx
```diff
@@ -0,0 +1,18 @@
+import { ministriesNames } from '@/ches/emailConstant';
+
+function checkUserRolesForMinistry(roles: string[]) {
+  return roles.some((role) => role.includes('ministry'));
+}
+
+export default function checkUserMinistryRole(roles: string[]): string | null {
```
junminahn, line 7: Let's revise the function to return `array of ministry names` as the user can have privileges over multiple ministries.

junminahn (commented):
I think it is getting closer. In general comments besides the individual comments on lines:
1. There is a `build` issues; please run `npm run build` to see what is breaking.
2. I understand that `getPrivateCloudProjectsQuery` changed to `async` function; then let's find its usages to add `await` to get the result.
3. Let's revise `checkUserMinistryRole` function to return array of ministry names and update the query accordingly to takes array of ministry names.

PR #1364, opened 2023-11-17
feat: add global api response handler

junminahn:
- since the api endpoints are captured errors and using `try.. catch` can be duplicated in multiple places, let's have a global handler to capture the responses and errors.


okanji (approved):
Looks good!

PR #1399, opened 2023-11-23
Fix nats

okanji:
Did some refactoring:

- Fixed nats messages. The function to format the nats  message was an async function which was not intended. It did not need to be async and was not being called as an async function.
- Combined comment modal with create modal. The comment modal still needed to have the text from the create modal. I have combined them into a single Edit modal.

image


junminahn (approved):
👍

PR #1440, opened 2023-11-29
Table enhancement 

okanji:
Edited the table to make it look better. This includes
- Adding avatars
- Adding spacing between columns
- Truncating description text so it does not go beyond two lines

image


junminahn (approved):
👍

PR #1442, opened 2023-11-29
Added filter select components

okanji:
- Converted the filter checkboxes to select components
- Made the `handleFilterChange` function general for any filter
- Create a `clearFilter` function to clear the url search params and re set the select components
- Created ministry and cluster constants from the Prisma Schema
- Updated the filter component to use the ministry and cluster constants
- Updated the forms to use the ministry and cluster constants

image


components/table/FilterPanel.tsx
```diff
@@ -1,56 +1,100 @@
...
+              id="ministry"
+              name="ministry"
+              autoComplete="cluster-name"
+              className="block w-full rounded-md border-0 py-1.5 text-gray-900 shadow-sm ring-1 ring-inset ring-gray-300 focus:ring-2 focus:ring-inset focus:ring-indigo-600 sm:max-w-xs sm:text-sm sm:leading-6"
+              onChange={(e) => handleFilterChange('ministry', e.target.value)}
+            >
+              <option selected={true} disabled value="">
+                Select Ministry
+              </option>
+              {ministries.map((cluster) => (
```
junminahn, line 82: is `cluster` typo? `cluster` > `ministry`?

constants/index.ts
```diff
@@ -0,0 +1,7 @@
+import { Cluster, Ministry } from '@prisma/client';
+
+export const clusters = Object.keys(Cluster)
```
junminahn, line 3: is
```
Object.keys(Cluster).map((key) => Cluster[key as keyof typeof Cluster])
```
the same as `Object.values(Cluster)`?

constants/index.ts
```diff
@@ -0,0 +1,7 @@
...
+  .map((key) => Cluster[key as keyof typeof Cluster])
+  .filter((cluster) => cluster !== 'GOLDDR');
+
+export const ministries = Object.keys(Ministry).map((key) => Ministry[key as keyof typeof Ministry]);
```
junminahn, line 7: is
```
Object.keys(Ministry).map((key) => Ministry[key as keyof typeof Ministry]);
```
the same as `Object.values(Ministry)`?

junminahn (approved):

PR #1443, opened 2023-11-30
Formatted table

okanji:
- Increased font size of table elements
- Changed avatars to only have the name. This makes the component shorter to save space in the table
image


junminahn (approved):
👍

PR #1474, opened 2023-11-30
Added public cloud emails folder template

okanji:
I have added a public cloud and private cloud folders for our emails. I have simply populated the public cloud emails by copy pasting the private cloud email templates.


junminahn (approved):
👍

PR #1505, opened 2023-12-05
delete email templates added

Kolezhanchik:
delete request, delete complete, delete rejected email templates added 
image
image


calbertan (approved):

ches/emailHandler.ts
```diff
@@ -93,3 +98,45 @@ export const sendEditRequestEmails = async (
...
+      body: email,
+      to: [product.projectOwner.email, product.primaryTechnicalLead.email, product.secondaryTechnicalLead?.email],
+      subject: `${product.name} deletion request has been approved`,
+    });
+  } catch (error) {
+    console.error('ERROR SENDING REQUEST REJECTION EMAIL');
+  }
+};
+
+export const sendDeleteRejectionEmails = async (product: PrivateCloudRequestedProjectWithContacts, comment: string) => {
```
calbertan, line 130: should be sendRejectionEmails. Can also add the sendDeleteRequest emails in /api/rivate-cloud/delete

Kolezhanchik: sendRejectionEmails gets request as an argument and than productName: request.requestedProject.name, while sendDeleteRejectionEmails  gets product as an argument and than productName: product.name

calbertan (dismissed):

ches/emailHandler.ts
calbertan: Would it be better if we make it just 1 function that takes string instead of product/request. So we just past the string as a parameter.

ches/emailHandler.ts
```diff
@@ -93,3 +98,45 @@ export const sendEditRequestEmails = async (
...
+  const email = render(DeleteApprovalTemplate({ product }), { pretty: true });
+
+  try {
+    await sendEmail({
+      body: email,
+      to: [product.projectOwner.email, product.primaryTechnicalLead.email, product.secondaryTechnicalLead?.email],
+      subject: `${product.name} deletion request has been approved`,
+    });
+  } catch (error) {
+    console.error('ERROR SENDING REQUEST REJECTION EMAIL');
```
calbertan, line 126: The error message for the send delete emails need to be updated

ches/emailHandler.ts
Kolezhanchik: it is impossible to pass only product name it also needs PO/TL emails

ches/emailHandler.ts
Kolezhanchik: fixed

calbertan (approved):

PR #1526, opened 2023-12-07
1466

calbertan:
Toggle button for Products
image
image

Toggle button for Requests
image
image

The same also exists for Public Cloud

### DIFF
--- a/queries/private-cloud/helpers.ts
+++ b/queries/private-cloud/helpers.ts
@@ -16,10 +16,10 @@ export async function getPrivateCloudProjectsQuery({
   ministryRoles?: string[];
   active?: string | null;
 }) {
-  const setActive = active === 'false' ? false : true;
+  const isActive = active === 'false' ? false : true;
 
   // Initialize the search/filter query
-  const searchQuery: any = setActive ? { status: 'ACTIVE' } : {};
+  const searchQuery: any = isActive ? { status: 'ACTIVE' } : {};
 
   // Construct search/filter conditions based on provided parameters
   if (searchTerm) {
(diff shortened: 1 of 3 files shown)


app/(dashboard)/private-cloud/products/page.tsx
```diff
@@ -52,6 +53,7 @@ export default async function ProductsTable({
     cluster,
     userEmail,
     ministryRoles,
+    JSON.parse(active || 'true'),
```
okanji, line 56: Do we need to do JSON.parse here? I believe active will be a string anyways but please verify

okanji (changes requested):

app/(dashboard)/private-cloud/products/page.tsx
calbertan: I wanted to pass a boolean since it makes more sense than passing a string. Another option is to pass the string, then do the parse in the function

okanji: I see, ok sounds good!

junminahn: Allow me to interject in the review process. It might introduce `errors (JSON parsing error)` if the query parameter `active` is anything other than `true` or `false`. It would be more reliable to directly compare the string values:
1.To set `false` as the default:
- `active === 'true'`

2. To set `true` as the default:
- `active !== 'false'`

queries/private-cloud/helpers.ts
```diff
@@ -7,17 +7,19 @@ export async function getPrivateCloudProjectsQuery({
...
+  active,
 }: {
   searchTerm?: string | null;
   ministry?: string | null;
   cluster?: string | string[] | null;
   userEmail?: string | null;
   ministryRoles?: string[];
+  active?: string | null;
 }) {
+  const setActive = active === 'false' ? false : true;
```
junminahn, line 19: `setActive` sounds more like a `method` rather than a variable.
something like `isActive` might be better name? 🤔

junminahn (approved):

PR #1537, opened 2023-12-07
Feat/1453

Kolezhanchik:
created and tested endpoints for aws user roles

### DIFF
--- a/app/api/public-cloud/aws-roles/routes.ts
+++ b/app/api/public-cloud/aws-roles/routes.ts
@@ -18,7 +18,12 @@ interface User {
   email: string;
 }
 
-async function getToken(): Promise<string | undefined> {
+const parseError = (error: unknown): void => {
+  if (error instanceof Error) {
+    console.log(error.message);
+  } else console.log(String(error));
+};
+const getToken = async (): Promise<string | undefined> => {
   try {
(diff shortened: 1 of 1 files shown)


Kolezhanchik:
done

app/api/public-cloud/aws-roles/routes.ts
```diff
@@ -0,0 +1,155 @@
...
+      },
+    });
+    return response.data.access_token;
+  } catch (error: unknown) {
+    parseError(error);
+  }
+};
+
+const awsRolesApiInstance = axios.create({
+  baseURL: `${process.env.AWS_ROLES_BASE_URL}/admin/realms/${process.env.AWS_ROLES_REALM_NAME}`,
```
junminahn, line 46: I see the baseUrl is different than one used for `getToken`:
- `${process.env.AWS_ROLES_BASE_URL}/admin/realms/${process.env.AWS_ROLES_REALM_NAME}`
vs
- `${process.env.AWS_ROLES_BASE_URL}/realms/${process.env.AWS_ROLES_REALM_NAME}`
can you double check which one is the correct base url and also we might define it on top as a variable to avoid mis-sync.

junminahn (commented):
just one more thing :)

app/api/public-cloud/aws-roles/routes.ts
Kolezhanchik: double checked, it is correct, they are different for get token and others

junminahn (dismissed):
👍

junminahn (approved):

PR #1557, opened 2023-12-12
filter for public clo9ud provider

Kolezhanchik:
filter for public cloud with provider, not cluster 
image
image

### DIFF
--- a/components/table/FilterPanel.tsx
+++ b/components/table/FilterPanel.tsx
@@ -2,6 +2,10 @@ import { usePathname, useRouter, useSearchParams } from 'next/navigation';
 import { clusters, ministries, providers } from '@/constants';
 import { useRef, useState } from 'react';
 
+const capitalizeFirstLetter = (name: string): string => {
+  return name.charAt(0).toUpperCase() + name.slice(1);
+};
+
 export default function FilterPanel() {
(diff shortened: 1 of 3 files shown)


components/table/FilterPanel.tsx
```diff
@@ -1,15 +1,16 @@
...
 export default function FilterPanel() {
   const [showInactive, setShowInactive] = useState(false);
   const { replace } = useRouter();
   const pathname = usePathname();
   const searchParams = useSearchParams()!;
-  const clusterRef = useRef<HTMLSelectElement>(null);
+  const clusterProviderRef = useRef<HTMLSelectElement>(null);
   const ministryRef = useRef<HTMLSelectElement>(null);
-
+  const currentClusterProvider = pathname.includes('private') ? 'Cluster' : 'Provider';
```
junminahn, line 12: As the `lowercase` string value is more used below, we could consider initially setting the value in `lowercase` and `capitalize` it when displaying it on the UI.
```js
function capitalizeFirstLetter(string) {
    return string.charAt(0).toUpperCase() + string.slice(1);
}
```

Kolezhanchik: changed, thanks

junminahn (approved):

PR #1608, opened 2023-12-19
Misc bug fixes

okanji:
- Fixed all redirect URLS
- Open secondary tech lead when it exists by default
- Fixed account coding
- Fixed budget
- Added correct calculator link
- Removed tool tip icons


app/(request)/private-cloud/edit/[licencePlate]/page.tsx
```diff
@@ -93,6 +93,18 @@ export default function EditProject({ params }: { params: { licencePlate: string
     }
   }, [requestData]);
 
+  useEffect(() => {
+    if (data?.secondaryTechnicalLead) {
+      setSecondTechLead(true);
+    }
+  }, [data]);
+
+  useEffect(() => {
```
Comment (junminahn) [inline app/(request)/private-cloud/edit/[licencePlate]/page.tsx:102]: is this a `duplication` of above?

Comment (okanji) [reply app/(request)/private-cloud/edit/[licencePlate]/page.tsx:102]: Yes but they are in different pages. So we want the secondary technical lead to be expanded in each of those pages if it exists

PR #1616, opened 2023-12-20
filtered LAB clusters for regular users

Kolezhanchik:
removed KLAB, KLAB2 and CLAB clusters for regular user role

### DIFF
--- a/components/form/ProjectDescriptionPrivate.tsx
+++ b/components/form/ProjectDescriptionPrivate.tsx
@@ -1,7 +1,8 @@
 import { useFormContext } from 'react-hook-form';
 import classNames from '@/components/utils/classnames';
-import { useRouter } from 'next/navigation';
 import { clusters, ministries } from '@/constants';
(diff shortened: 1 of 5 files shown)


components/form/ProjectDescriptionPrivate.tsx
```diff
@@ -17,7 +18,17 @@ export default function ProjectDescription({
...
 
-  const router = useRouter();
+  const [clustersList, setClustersList] = useState(clusters);
+
+  const { data: session } = useSession({
+    required: true,
+  });
+
+  useEffect(() => {
+    if (session && !session?.isAdmin) {
```
junminahn, line 28: I think `session &&` is redundant as we use `?.isAdmin`?

junminahn (approved):

components/form/ProjectDescriptionPrivate.tsx
Kolezhanchik: removed, thanks

PR #1624, opened 2023-12-21
feat: add zap vulnerability scanning airflow DAG

junminahn:


security-dashboard/airflow/Dockerfile
Amritpal-Nijjar, line 1: Just wondering if this a concern at all ?
Screenshot 2023-12-21 at 1 07 56 PM

junminahn: @Amritpal-Nijjar glad that you checked the `sonarQube` checks. as it is based on the official `airflow image` and also `Openshift` overrides the container user to `nonroot` users, so there is not concern of running the container as `root` user on Openshift.

Amritpal-Nijjar: Good to know!

Amritpal-Nijjar (approved):

PR #1630, opened 2023-12-22
chore: ignore NEXTAUTH_URL to allow multi-domains

junminahn:
- rename `.env.localdev` to `.env.example`
- ignore `NEXTAUTH_URL` to allow multi-domains


Amritpal-Nijjar (approved):
🚀

PR #1702, opened 2024-01-08
delete functionality implemented

Kolezhanchik:
added delete functionality to aws users dashbaord


app/api/public-cloud/aws-roles/deleteUser/route.ts
```diff
@@ -0,0 +1,34 @@
+import { removeUserFromGroup } from '@/app/api/public-cloud/aws-roles/route';
+import { NextRequest, NextResponse } from 'next/server';
+
+export async function DELETE(req: NextRequest, res: NextResponse) {
```
junminahn, line 4: Would you try applying the `api handler` for the new endpoints?
I believe it would simplify the codebase:
https://github.com/bcgov/platform-services-registry/blob/main/app/api/example/%5Bcolor%5D/route.ts

junminahn (commented):
In general, is there any chance to re-use the same route for 
`app/(product)/public-cloud/(product)/aws-roles/[licencePlate]/[role]/page.tsx`?
I believe `role` segment also can be passed as `path param` like `licensePlate`.

app/api/public-cloud/aws-roles/deleteUser/route.ts
Kolezhanchik: changes are made, thank you!

junminahn (approved):
👍

PR #1767, opened 2024-01-15
Analytics

okanji:
# Created analytics page

image


analytics/private-cloud/quotaChanges.ts
```diff
@@ -0,0 +1,95 @@
...
+  return formatter.format(date);
+}
+
+export function isQuotaChanged(
+  projectOne: PrivateCloudRequestedProject,
+  projectTwo: PrivateCloudRequestedProject,
+): boolean {
+  // Assuming productionQuota, testQuota, developmentQuota, and toolsQuota are defined and comparable
+  return !(
+    JSON.stringify(projectOne.productionQuota) === JSON.stringify(projectTwo.productionQuota) &&
```
junminahn, line 20: maybe we can try `lodash` `isEqual` function to compare two objects?
```sh
import _isEqual from 'lodash-es/isEqual';
_isEqual(projectOne.productionQuota, projectTwo.productionQuota)
```

Note that I forgot to install `types` when installing `lodash-es`, please install it
```sh
npm i --save-dev @types/lodash-es
```

analytics/private-cloud/quotaChanges.ts
```diff
@@ -0,0 +1,95 @@
...
+    monthlyChanges[monthKey].push(date);
+  });
+  return monthlyChanges;
+}
+
+export async function detectQuotaChangesByMonth(licencePlate: string, decisionStatus?: DecisionStatus) {
+  const requests = await prisma.privateCloudRequest.findMany({
+    where: { licencePlate, decisionStatus },
+    orderBy: { created: 'asc' },
+
```
junminahn, line 43: if `requestedProject` is the only field required for the logic, we can add it to the query like:
```
{
...
select: { requestedProject: true },
...
}

File: analytics/private-cloud/quotaChanges.ts [hunk 3]
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  const monthlyChanges = sortDatesIntoMonths(quotaChangeDates);
+  return monthlyChanges;
+}
+
+export async function quotaEditRequests(decisionStatus?: DecisionStatus) {
+  const projects = await prisma.privateCloudProject.findMany();
+
+  if (!projects) {
```
junminahn, line 66: does `findMany` returns `falsy` value if empty?
if it returns empty array `[]`, then the condition won't match as expected.

analytics/private-cloud/quotaChanges.ts
```diff
@@ -0,0 +1,95 @@
...
+    throw new Error('No projects found in the database');
+  }
+
+  const licencePlates = projects.map((project) => project.licencePlate);
```
junminahn, line 70: if `licencePlate` is the only field we need from the collection, we can add it to `select` cause of the query.

app/api/private-cloud/analytics/quota-requests/csv/route.ts
```diff
@@ -0,0 +1,51 @@
...
+};
+
+const apiHandler = createApiHandler({
+  roles: ['admin'],
+});
+
+export const GET = apiHandler(async () => {
+  const allData = await quotaEditRequests();
+  const approvedData = await quotaEditRequests(DecisionStatus.APPROVED);
+  const rejectedData = await quotaEditRequests(DecisionStatus.REJECTED);
```
junminahn, line 21: We could group them all into `Promise.all` to reduce the wait time.

okanji:
Great suggestions, I will integrate all of them.

junminahn (approved):

PR #1816, opened 2024-01-18
New analytics

okanji:
- Made analytics page available to all users
- Added page title
- Added chart subtitles
- Added api endpoint for histogram csv


junminahn (commented):
If the `roles` parameter is empty in the `API handler`, it grants access to the page for any guest users. To restrict access to authenticated users with IDIR, include the `user` role within the brackets.

okanji:
> If the `roles` parameter is empty in the `API handler`, it grants access to the page for any guest users. To restrict access to authenticated users with IDIR, include the `user` role within the brackets.

Got it! Thanks

junminahn (approved):

PR #1820, opened 2024-01-18
Feat/1731

Kolezhanchik:
search functionality by substring for aws roles users


services/aws-roles.ts
```diff
@@ -1,9 +1,15 @@
 import { instance } from './axios';
 
-export async function GetUsersPaginatedList(licencePlate: string, role: string, currentPage: string, pageSize: string) {
+export async function GetUsersPaginatedList(
```
junminahn, line 3: I believe we have naming condition for JS function name as `camelCase`?

app/api/public-cloud/aws-roles/helpers.ts
```diff
@@ -40,6 +41,44 @@ const paginate = <T>(users: T[], options: PaginationOptions): T[] => {
   return users.slice(startIndex, endIndex);
 };
 
+const searchSubstringInArray = (searchTerm: string, users: Record<string, User>[]): Record<string, User>[] => {
+  const results = new Set<Record<string, User>>();
+
+  users.filter((userObj) =>
```
junminahn, line 47: Hmm, I'm curious about the data structure of 'users' since it involves nested arrays. In general, when the purpose is iterating over arrays and the 'filter' and 'some' functions are used to produce output, replacing them with 'forEach' can help avoid confusion.

app/api/public-cloud/aws-roles/helpers.ts
```diff
@@ -40,6 +41,44 @@ const paginate = <T>(users: T[], options: PaginationOptions): T[] => {
...
+
+  for (const key in data) {
+    if (key in newUser) {
+      newUser[key as keyof V] = data[key] as V[keyof V];
+    }
+  }
+  return newUser;
+};
+
+const userRole = <K extends string, V extends User>(role: K, user: V): Record<string, V> => {
```
junminahn, line 77: maybe `createUserRole` could be a better name for the function?

app/api/public-cloud/aws-roles/helpers.ts
```diff
@@ -56,17 +95,19 @@ async function getPublicCloudProjectUsers(searchLicencePlate: string): Promise<R
 
   if (project?.projectOwnerId) {
     const projectOwner = await getUserById(project.projectOwnerId);
```
junminahn, line 97: I know, it is reviewed already before, but can we use `include` clause of `prisma` query to avoid the additional database trips?

junminahn (commented):
For the user table, I see it truncates the users after pulling all users from Keycloak and our database, I am concerned of its performance once the number of the users grow eventually.

services/aws-roles.ts
Kolezhanchik: sure,  changed

app/api/public-cloud/aws-roles/helpers.ts
Kolezhanchik: yes, forEach works perfectly, thanks

app/api/public-cloud/aws-roles/helpers.ts
Kolezhanchik: changed

app/api/public-cloud/aws-roles/helpers.ts
Kolezhanchik: yes, that is much better, changed

junminahn (approved):

PR #1850, opened 2024-01-22
added routing for Nav

Kolezhanchik:


components/nav/Nav.tsx
```diff
@@ -2,16 +2,29 @@ import Image from 'next/image';
 import Logo from '@/components/assets/logo.svg';
 import ProfileDropdown from '@/components/nav/ProfileDropdown';
 import Link from 'next/link';
+import { headers } from 'next/headers';
+
+export default async function NavBar() {
+  const headersList = headers();
+  const referer = headersList.get('referer');
+  let url = '/private-cloud/products/all';
+  if (typeof referer === 'string') {
```
junminahn, line 11: let's use `import _isString from 'lodash-es/isString` for better readability.

components/nav/Nav.tsx
```diff
@@ -2,16 +2,29 @@ import Image from 'next/image';
...
+    if (referer.split('/')[3] === 'public-cloud') {
```
junminahn, line 12: let's create a variable on top to reuse the variable for two conditions.

components/nav/Nav.tsx
```diff
@@ -2,16 +2,29 @@ import Image from 'next/image';
...
+      url = '/public-cloud/products/all';
+    }
+    if (referer.split('/')[3] === 'private-cloud') {
```
junminahn, line 15: let's use `else if`.

components/nav/Nav.tsx
Kolezhanchik: currentURL variable is created

components/nav/Nav.tsx
Kolezhanchik: changed to using _isString

components/nav/Nav.tsx
Kolezhanchik: doesn't work properly with `else if`

junminahn: hm.. not sure why though. can you explain little more?

Kolezhanchik: also changed, thank you

junminahn (approved):

PR #1871, opened 2024-01-24
product page tabs fixies

Kolezhanchik:


components/generic/Tabs.tsx
```diff
@@ -38,7 +38,7 @@ export default function Tabs({ tabs, children }: { tabs: ITab[]; children: React
             <div className=" -mb-px flex justify-start">
               {tabs.map((tab) => (
                 <Link
-                  href={tab.href}
```
junminahn, line 41: can you try without the function `pathChangeTail`? it just works for me.

Kolezhanchik: without `pathChangeTail` user pushes tabs `ROLES` being on e.g. `Developers` or any other  and users is pushed to `Admins`

junminahn: @Kolezhanchik sorry, not following you comment, would you mind give me little more details with actual URLs?

Kolezhanchik: when user pushes tab `ROLES` on product page tabs user is routed to 
image
/public-cloud/aws-roles/eu9cfk/developers

user being in tab  `ROLES` pushes tab `Developers` and is routed to /public-cloud/aws-roles/eu9cfk/developers
image

being in `ROLES` + `Developers` user can push `ROLES`(which is clickable)  and this user is routed to `Admins`, which is not expected(expected to stay on `Developers` tab

junminahn: @Kolezhanchik I think I am getting it and we need to identify that it is actually a bug though. I see the ticket you are working on and it mentions of the `ROLES tab not highlighted` as the bug.
So, do you think it is a bug? to me it is an expected behaviour as the `Admins` tab is the default tab of the upper level `ROLES` tab.

Kolezhanchik: removed `pathChangeTail` from Tabs

junminahn (approved):

PR #1924, opened 2024-01-30
feat: add comments model for prisma schema w/ reactions

Amritpal-Nijjar:
To create the preliminary ground work for a comments feature. Two models created: 1)Comment 2) Reaction.


junminahn (commented):
I believe we would need to have a discussion to create the DB schema.

Amritpal-Nijjar:
> I believe we would need to have a discussion to create the DB schema.

I messed up, this is supposed to be merged into feat/comments branch not the main branch, will redo. Sorry for the confusion.

PR #1955, opened 2024-02-01
Public cloud analytics

okanji:
Created the public cloud analytics page similar to the private cloud


analytics/public-cloud/requests.ts
```diff
@@ -0,0 +1,96 @@
...
+    },
+    select: {
+      created: true,
+    },
+    orderBy: {
+      created: 'asc',
+    },
+  });
+
+  const deleteRequests = await prisma.publicCloudRequest.findMany({
```
junminahn, line 54: I believe we can use `Promise.all` to get the performance gain by sending multiple DB calls concurrently.

okanji: It becomes messy to do it with Promise.all:

junminahn (approved):

PR #1999, opened 2024-02-07
fix(1680): fix comments in emails

Amritpal-Nijjar:
This PR has fixed the issue of `comments` not appearing in emails when an `edit`, `approval`, or `rejection` is made in both Public and Private Clouds. All email subjects are now consistent and generic. Refactored the schema in Public cloud to now be `adminComment` instead of `humanComment`

### DIFF
--- a/emails/_components/Comment.tsx
+++ b/emails/_components/Comment.tsx
@@ -1,0 +1,24 @@
+import { Text } from '@react-email/components';
+
+interface CommentProps {
+  userComment?: string | null;
+  adminComment?: string | null;
+}
+
+export default function Comment({ userComment, adminComment }: CommentProps) {
+  return (
+    <div>
+      {userComment && (
+        <div>
+          <Text>{userComment}</Text>
+        </div>
+      )}
+      {adminComment && (
+        <div>
+          <Text className="font-semibold">Admin Comment:</Text>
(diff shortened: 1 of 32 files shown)


emails/_components/Comment.tsx
```diff
@@ -0,0 +1,24 @@
...
+  userComment?: string | null;
+  adminComment?: string | null;
+}
+
+export default function Comment({ userComment, adminComment }: CommentProps) {
+  return (
+    <div>
+      {userComment && (
+        <div>
+          <Text>{userComment}</Text>
```
junminahn, line 13: curious if we also need title for `User Comment:` as well just like we have the `Admin Comment` below.

Amritpal-Nijjar: I was a little unsure on what to do in this case, especially since `userComment` is in the schema for Private Cloud, but not in Public Cloud, and just wanted to have extensibility with this component. But then styling of the Private Cloud and Public Cloud would differ... So, I left out the title/heading for `userComment`

junminahn: it is more for `UI design` rather than the schema changes; I will approve it now and let's see.

junminahn (approved):

PR #2038, opened 2024-02-12
prevent page reboot on search

Kolezhanchik:
prevent page reboot on hitting "Enter" button in search input field

### DIFF
--- a/components/table/SearchFilterSort.tsx
+++ b/components/table/SearchFilterSort.tsx
@@ -106,7 +106,6 @@ export default function SearchFilterSort({
         <div className="flex flex-grow-0 justify-end space-x-2.5 w-full items-center">
           <div className="flex w-full justify-between items-center">
             <div className="flex-grow h-12"></div>
-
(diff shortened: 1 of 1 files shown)


components/table/SearchFilterSort.tsx
```diff
@@ -134,6 +133,10 @@ export default function SearchFilterSort({
                     onBlur={() => setFocused(false)}
                     onChange={(e) => setSearchTerm(e.target.value)}
                     spellCheck={false}
+                    onKeyDown={(e) => {
+                      //prevent page reboot on "Enter" key press
```
junminahn, line 137: I believe we have a comment convention to start with empty character followed by a capital word;
so can we change it to `// Prevent page reboot on "Enter" key press`?
I will look for a `eslint` plugin to set it as a eslint rules.

Kolezhanchik: done

junminahn (approved):

PR #2086, opened 2024-02-17
fix: return modal + hosting tier info for users

Amritpal-Nijjar:


components/modal/Return.tsx
```diff
@@ -31,6 +33,27 @@ export default function Modal({
     router.refresh();
   };
 
+  // Dynamically setting the modal title
+  let modalTitle = 'Thank you! We have received your ';
```
junminahn, line 37: instead of have conditions to set title & message, can you have them as `arguments` of the component?

junminahn (approved):

PR #2527, opened 2024-04-22
Scenario #4 - Emails refactoring/improving according to docs (Private Cloud)

Amritpal-Nijjar:
1. Admins now receive an email notifying them of a pending delete request that needs review
2. Updated the documentation and the content of emails to be consistent
3. Comments box does not appear for when approving a delete request
4. Sending out a deleteApproval email after delete provisioning has been completed


app/api/private-cloud/provision/[licencePlate]/route.ts
```diff
@@ -78,5 +78,9 @@ export const PUT = apiHandler(async ({ pathParams }) => {
   if (request.type == 'CREATE') {
     await wrapAsync(() => sendProvisionedEmails(project as PrivateCloudRequestedProjectWithContacts));
   }
+
+  if (request.type == 'DELETE') {
```
Comment (junminahn) [inline app/api/private-cloud/provision/[licencePlate]/route.ts:82]: we can use `else if` here.

junminahn (approved):

PR #2562, opened 2024-04-24
chore: route files after review fixies

Kolezhanchik:


app/api/private-cloud/products/[licencePlate]/history/route.ts
```diff
@@ -25,8 +25,8 @@ export const GET = apiHandler(async ({ pathParams, session }) => {
     session: session as never,
   });
 
-  if (!requests) {
-    return NotFoundResponse('No requests found for this licence plate.');
+  if (requests.length < 1) {
```
Comment (junminahn) [inline app/api/private-cloud/products/[licencePlate]/history/route.ts:28]: I don't think an empty array is `no content`. an empty array is also valid result so that we can return it as it is.
also, the route should be `app/api/private-cloud/products/[licencePlate]/requests/route.ts`

app/api/public-cloud/history/[licencePlate]/route.ts
```diff
@@ -1,7 +1,7 @@
 import prisma from '@/core/prisma';
```
Comment (junminahn) [inline app/api/public-cloud/history/[licencePlate]/route.ts:1]: I believe this route is no longer required?

app/api/public-cloud/requested-project/[id]/route.ts
```diff
@@ -12,7 +12,7 @@ const apiHandler = createApiHandler({
   validations: { pathParams: pathParamSchema },
 });
 
-export const GET = apiHandler(async ({ pathParams }) => {
```
Comment (junminahn) [inline app/api/public-cloud/requested-project/[id]/route.ts:15]: this route will return no data as `core service` does not support for the schema `requested project`.
let's remove this route and rename `app/api/private-cloud/request/[id]` to `app/api/private-cloud/requests/[id]` to fetch requested projects for a request.

app/api/public-cloud/history/[licencePlate]/route.ts
Comment (Kolezhanchik) [reply app/api/public-cloud/history/[licencePlate]/route.ts:1]: if this route is no longer required, does it mean that it should be `app/api/public-cloud/products/[licencePlate]/requests/route.ts` ?

PR #2632, opened 2024-04-30
chore: let admin, cloud-admins and ministry-editors see history tab

Kolezhanchik:


app/api/private-cloud/products/[licencePlate]/requests/route.ts
```diff
@@ -16,6 +17,7 @@ const queryParamSchema = z.object({
 
 const apiHandler = createApiHandler({
   roles: ['user'],
+  permissions: [PermissionsEnum.ViewAllPrivateCloudProductsHistory],
```
Comment (junminahn) [inline app/api/private-cloud/products/[licencePlate]/requests/route.ts:20]: please consider removing this permission; this api endpoint is for all users and I will get data what they can get only based on the security layer.

app/api/public-cloud/products/[licencePlate]/requests/route.ts
```diff
@@ -16,6 +17,7 @@ const queryParamSchema = z.object({
 
 const apiHandler = createApiHandler({
   roles: ['user'],
+  permissions: [PermissionsEnum.ViewAllPublicCloudProductsHistory],
```
Comment (junminahn) [inline app/api/public-cloud/products/[licencePlate]/requests/route.ts:20]: please consider removing this permission; this api endpoint is for all users and I will get data what they can get only based on the security layer.

PR #2682, opened 2024-05-06
Feat/2566 add script to populate originalDataId renamed requestedProjectId and userRequestedProjectI

Kolezhanchik:


data-migrations/migrations/20240502202804-add_original_data_id_fields_private.js
```diff
@@ -0,0 +1,29 @@
...
+
+  const privateCloudRequests = await PrivateCloudRequest.find({ originalDataId: { $exists: false } }).toArray();
+
+  for (const privateCloudRequest of privateCloudRequests) {
+    const licencePlateVal = privateCloudRequest.licencePlate;
+    const privateCloudRequestCreated = new Date(privateCloudRequest.created.toISOString());
+
+    const matchingPrivateCloudRequestedProjects = await PrivateCloudRequestedProject.find({
+      licencePlate: licencePlateVal,
+      created: { $lte: privateCloudRequestCreated },
```
junminahn, line 13: I don't think `created` field is reliable field to query onto.

data-migrations/migrations/20240502202804-add_original_data_id_fields_private.js
```diff
@@ -0,0 +1,29 @@
...
+    })
+      .sort({ created: -1 })
+      .toArray();
+
+    if (matchingPrivateCloudRequestedProjects) {
```
junminahn, line 18: since we use `find` and `toArray` convert the results into an array, it could just an empty array rather than `null`.

junminahn (commented):
I believe there is no logic to set `originalDataId` for the request document.

data-migrations/migrations/20240502202804-add_original_data_id_fields_private.js
Kolezhanchik: matchingPrivateCloudRequestedProjects will be in array with the same `licencePlate` and only `created` is different

junminahn: let's discuss this in another PR. this PR is too big.

junminahn (approved):

PR #2792, opened 2024-05-14
chore: revisit populate original data script

Kolezhanchik:


data-migrations/migrations/20240514162736-populate_origin_data_field_public.js.js
```diff
@@ -0,0 +1,46 @@
...
+          },
+        },
+      ]).toArray();
+
+      for (const requestsLicencePlate of publicCloudRequests) {
+        if (requestsLicencePlate.totalAmount > 1) {
+          const publicCloudRequestsLicencePlate = await PublicCloudRequest.find({
+            licencePlate: requestsLicencePlate._id,
+          })
+            .sort({ created: -1 })
```
junminahn, line 22: I believe `1` is `ASC`; means `old to new`.

Kolezhanchik: actually, we need the last one(`newest one`) request first, to assign it previous request `decisionDataId`. If it would be `old to new`, than we would need to try to the oldest request to assign next(`newer`) request `decisionDataId`

junminahn (approved):

PR #2986, opened 2024-05-31
Feat/2919 a

Kolezhanchik:
rename `TEST` to `TEMP`,  change wording to `XX days overdue for automatic deletion`, remove ability to change `isTest` status for edit page, wording for toggle filter Test Products, show Temp in dashboard as label for products and requests and popup only, disable the option to unselect once provisioned


app/app/public-cloud/requests/(request)/[id]/layout.tsx
```diff
@@ -13,9 +13,24 @@ import LightButton from '@/components/generic/button/LightButton';
 import Tabs, { ITab } from '@/components/generic/tabs/BasicTabs';
 import createClientPage from '@/core/client-page';
 import { comparePublicProductData } from '@/helpers/product';
+import { PublicCloudRequestGetPayload } from '@/queries/public-cloud-requests';
 import { getPublicCloudRequest } from '@/services/backend/public-cloud/requests';
 import { publicProductState } from '@/states/global';
 
+const mapToComponentData = (payload?: PublicCloudRequestGetPayload, isTest?: boolean) => {
```
Comment (junminahn) [inline app/app/public-cloud/requests/(request)/[id]/layout.tsx:20]: hm.. not sure why we need this mapper as `decision data` should have the flag `isTest`.

app/components/form/RequestBadge.tsx
```diff
@@ -3,11 +3,21 @@ import { $Enums } from '@prisma/client';
...
+interface DecisionData {
+  isTest?: boolean;
+}
+
 export default function RequestBadge({
   data,
   className,
 }: {
-  data?: { licencePlate: string; type: $Enums.RequestType; active: boolean; decisionStatus: $Enums.DecisionStatus };
+  data?: {
```
junminahn, line 14: let's re-organize the props so to have a single prop `request` which will have all information needed.

app/components/form/RequestBadge.tsx
```diff
@@ -54,6 +64,11 @@ export default function RequestBadge({
       <Badge color={decisionColor} radius="sm" className="ml-1">
         {data.decisionStatus}
       </Badge>
+      {data.decisionData?.isTest ? (
```
junminahn, line 67: let's use `&&` instead of `?` to get rid of `null` at the end.

app/components/table/TableBodyPrivateProducts.tsx
```diff
@@ -89,9 +91,7 @@ export default function TableBodyPrivateProducts({ rows, isLoading = false }: Ta
               {row.activeRequest && <ActiveRequestBox data={{ ...row.activeRequest, cloud: 'private-cloud' }} />}
             </div>
 
-            <div className="md:col-span-2 lg:col-span-2">
-              {row.isTest && <TestProductBox data={{ createdAt: row.createdAt }} />}
-            </div>
+            <div className="lg:col-span-1 hidden lg:block"></div>
```
junminahn, line 94: can we revisit the `grid & columns` to have the structure as before?

app/app/public-cloud/requests/(request)/[id]/layout.tsx
Comment (Kolezhanchik) [reply app/app/public-cloud/requests/(request)/[id]/layout.tsx:20]: public cloud product doesn't have `isTest` flag and this feature is not relevant for public cloud products

app/components/table/TableBodyPrivateProducts.tsx
Kolezhanchik: actually it is the structure "as before", copied from history

junminahn (approved):

PR #3155, opened 2024-06-20
fix(3093): add word wrap and character limit to description tool tips

Amritpal-Nijjar:
**Before**
Screenshot 2024-06-20 at 1 17 27 PM
**After**
Screenshot 2024-06-20 at 1 21 43 PM


junminahn (approved):
💯

PR #3177, opened 2024-06-21
fix: show only requested environments budget

Kolezhanchik:


app/emails/_components/ProviderDetails.tsx
```diff
@@ -7,16 +7,24 @@ type Budget = {
   tools: number;
 };
 
+type EnvironmentsEnabled = { production: boolean; test: boolean; development: boolean; tools: boolean };
```
junminahn, line 10: let's use the one from `prisma` generated types; 

```
import { EnvironmentsEnabled  } from '@prisma/client';
```

junminahn (approved):

PR #3193, opened 2024-06-24
feat: add subscription to mailchimp

Kolezhanchik:


Amritpal-Nijjar (approved):

Kolezhanchik:
> Shouldn't we also include a step to clear all emails from the list before adding the valid ones?

we have a step to remove tag(to remove an email from segment), an email in a list can have other tags, besides, it is better to archive an email than to remove it from mailchimp list

PR #3194, opened 2024-06-24
feat(3179): add rich text editor to admin comments pages

Amritpal-Nijjar:
Screenshot 2024-06-24 at 11 26 24 AM


app/components/comments/CommentBubble.tsx
```diff
@@ -40,6 +55,41 @@ const CommentBubble = ({
   const [editMode, setEditMode] = useState(false);
   const [editedText, setEditedText] = useState(text);
 
+  const editEditor = useEditor({
```
junminahn, line 58: Let's define the common configuration in a separate file so it can be used in multiple places but still be overridden if needed.

junminahn (approved):

PR #3201, opened 2024-06-24
feat(3145): integrate the provisioner script into Airflow DAGs

funtigr:
Add the tested in Airflow backend 'backup callback' logic and two DAGs, one for Test (tested in Airflow, triggered manually) and one for Prod

### DIFF
--- a/helm/secdash/dags/mark_provisioned.py
+++ b/helm/secdash/dags/mark_provisioned.py
@@ -1,10 +1,10 @@
-import os
 import requests
-from pymongo import MongoClient
 from bson.objectid import ObjectId
+from projects import get_mongo_db
+from _utils import keys_exist
 
 
-def fetch_products_mark_completed(provisioner_api_url, mark_provisioned_url, mongo_connection_url):
+def fetch_products_mark_completed(provisioner_api_url, mark_provisioned_url, mongo_conn_id):
     """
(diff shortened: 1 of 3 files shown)


helm/secdash/dags/mark_provisioned.py
```diff
@@ -40,7 +39,12 @@ def fetch_products_mark_completed(provisioner_api_url, mark_provisioned_url, mon
             workflow_name = f"{cluster}-{licence_plate}-{request_id}"
             provisioner_workflow_url = f"{provisioner_api_url}/{workflow_name}"
             request_status_in_provisioner = requests.get(provisioner_workflow_url).json()
-            request_phase = request_status_in_provisioner['metadata']['labels']['workflows.argoproj.io/phase']
+            if 'metadata' not in request_status_in_provisioner:
```
junminahn, line 42: if `metadata` does not exists, let's just exist the function.

funtigr: When I was testing, it printed an error message:
```
{
    "code": 5,
    "message": "workflows.argoproj.io \"klab-dc5249-666b6bbd9d3e7453bbad1e66\" not found"
}
```
 so I thought it would be useful for understanding what went wrong if metadata is not available.

junminahn (approved):
👍

PR #3222, opened 2024-06-26
feat: add support phone number for private products

Kolezhanchik:


app/components/form/TeamContacts.tsx
```diff
@@ -105,6 +126,32 @@ export default function TeamContacts({
...
+            className="w-fit"
+            mask="+1 (000) 000-0000"
+            unmask={true}
+            value={phoneNumber}
+            onAccept={handlePhoneNumberChange}
+            placeholder="+1 (999) 999-9999"
+          />
+          <p
+            className={classNames(
+              errors.supportPhoneNumber ? 'text-red-400' : '',
```
junminahn, line 148: I'm wondering if `text-red-400` will override `text-gray-600` even though it is defined before `text-gray-600`.

app/schema.ts
```diff
@@ -116,6 +118,21 @@ export const PrivateCloudCreateRequestBodySchema = z.object({
   commonComponents: CommonComponentsInputSchema,
   golddrEnabled: z.preprocess(processBoolean, z.boolean()),
   isTest: z.preprocess(processBoolean, z.boolean()),
+  supportPhoneNumber: z
+    .string()
+    .nullable()
```
junminahn, line 123: Are `nullable` and `optional` still necessary given that these values are handled within the `refine` method?

junminahn (approved):

PR #3236, opened 2024-06-26
feat(3207): display the number of comments on a request to admins

Amritpal-Nijjar:
Screenshot 2024-06-26 at 1 55 05 PM


junminahn (commented):
What if we add a `commentCount` field to the `list` and `read` routes of the comments? This would make the frontend much simpler and easier to maintain.

Amritpal-Nijjar:
> What if we add a `commentCount` field to the `list` and `read` routes of the comments? This would make the frontend much simpler and easier to maintain.

Good idea, I will do that instead!

junminahn (commented):
1. Add a backend endpoint `/api/private-cloud/products/[licencePlate]/count` with an optional query parameter `requestId`.
    - If `requestId` is specified, the endpoint should return the number of comments for that specific request. Otherwise, it should return the number of comments for the specific product.

2. Add a corresponding service on the frontend.

3. Add a prop called `showCount` to the `ActiveRequestBox` component to fetch the number of comments for the request using `@tanstack/react-query`.

app/app/api/private-cloud/products/[licencePlate]/count/_operations/count.ts
```diff
@@ -0,0 +1,25 @@
+import prisma from '@/core/prisma';
+
+export async function getCommentCountOp(licencePlate: string, requestId?: string) {
+  if (requestId) {
+    const commentCount = await prisma.privateCloudComment.count({
+      where: { requestId },
```
Comment (junminahn) [inline app/app/api/private-cloud/products/[licencePlate]/count/_operations/count.ts:6]: I know it might not be necessary, but let's add `projectId` to the filter to ensure its uniqueness.

app/app/api/private-cloud/products/[licencePlate]/count/route.ts
```diff
@@ -0,0 +1,28 @@
...
+  permissions: [PermissionsEnum.ViewAllPrivateProductComments],
+  validations: {
+    pathParams: pathParamsSchema,
+    queryParams: queryParamsSchema,
+  },
+})(async ({ pathParams, queryParams }) => {
+  const { licencePlate } = pathParams;
+  const { requestId } = queryParams;
+
+  const commentCount = await getCommentCountOp(licencePlate, requestId);
```
Comment (junminahn) [inline app/app/api/private-cloud/products/[licencePlate]/count/route.ts:26]: let's call `commentCount` to more generic term like `res`.

app/components/form/ActiveRequestBox.tsx
```diff
@@ -10,23 +10,40 @@ import {
...
 }) {
   const router = useRouter();
+  const { data: session } = useSession();
+
+  const canViewComments = session?.permissions?.viewAllPrivateProductComments;
+
+  const { data: commentData, isLoading } = useQuery({
+    queryKey: ['commentCount', data?.id],
+    queryFn: () => getCommentCount(data?.licencePlate as string, data?.id as string),
+    enabled: showCount && canViewComments && !!data?.id && !!data?.licencePlate,
```
junminahn, line 43: Let's encapsulate the entire statement as an argument so that the variable can be reused on line 142.

app/components/form/ActiveRequestBox.tsx
```diff
@@ -132,26 +131,37 @@ export default function ActiveRequestBox({
...
-          </Badge>
-          <div>{badges}</div>
-          {data.createdByEmail && (
-            <div className="text-center text-sm text-gray-400">
-              Submitted by <span>{data.createdByEmail}</span>
-            </div>
-          )}
-        </Indicator>
+        <div className="relative">
+          {showCount && canViewComments && !isLoading && typeof commentCount === 'number' && commentCount > 0 ? (
```
junminahn, line 142: see the comment above

app/components/table/TableBodyPrivateProducts.tsx
```diff
@@ -88,7 +89,12 @@ export default function TableBodyPrivateProducts({ rows, isLoading = false }: Ta
               </div>
             </div>
             <div className="md:col-span-2 lg:col-span-3">
-              {row.activeRequest && <ActiveRequestBox data={{ ...row.activeRequest, cloud: 'private-cloud' }} />}
+              {row.activeRequest && (
+                <ActiveRequestBox
+                  data={{ ...row.activeRequest, cloud: 'private-cloud', licencePlate: row.licencePlate }}
+                  showCount={true}
```
junminahn, line 95: Let's drop `{true}` since it's not necessary.

app/services/backend/private-cloud/products.ts
```diff
@@ -126,3 +126,9 @@ export async function deletePrivateCloudComment(licencePlate: string, commentId:
   const response = await instance.delete(`/${licencePlate}/comments/${commentId}`);
   return response.data as { success: boolean };
 }
+
+export async function getCommentCount(licencePlate: string, requestId?: string) {
```
junminahn, line 130: Let's name it `getPrivateCloudCommentCount` to follow the naming convention.

junminahn (commented):
in overall, looking great!

app/app/api/private-cloud/products/[licencePlate]/count/_operations/count.ts
Comment (Amritpal-Nijjar) [reply app/app/api/private-cloud/products/[licencePlate]/count/_operations/count.ts:6]: Only problem is that adding `projectId` to the filter will not work for `create requests` as they do not have a `projectId` until that create request has been approved...

junminahn (approved):

PR #3318, opened 2024-07-11
test(3164): add create op test for admin comments

Amritpal-Nijjar:


app/app/api/private-cloud/products/[licencePlate]/comments/_operations/create.ts
```diff
@@ -12,9 +14,16 @@ export async function createOp(text: string, userId: string, projectId?: string,
     requestId: requestId || undefined,
   };
 
-  const comment = await prisma.privateCloudComment.create({
-    data,
-  });
+  try {
```
Comment (junminahn) [inline app/app/api/private-cloud/products/[licencePlate]/comments/_operations/create.ts:17]: Let's avoid using `try` here so the error can be caught at a higher level as `5xx` error.

app/core/api-handler.ts
```diff
@@ -209,6 +210,9 @@ function createApiHandler<
 
         return await fn({ session, pathParams, queryParams, body, headers: req.headers, jwtData });
       } catch (error) {
+        if (error instanceof BadRequestError) {
```
junminahn, line 213: I don't think it's necessary.

junminahn (approved):

PR #3549, opened 2024-08-20
feat(3467): refactor create test for private and public cloud

funtigr:


junminahn (commented):
Overall, it looks good.

1. Let's use double quotes instead of single quotes. You can refer to this guide: - https://cucumber.io/docs/bdd/better-gherkin/

2. Let's include the double quotes in the regular expression to avoid the unnecessary use of the `removeQuotes` function.

junminahn (approved):

PR #3613, opened 2024-08-27
feat(3191): add admin tooltip and notification

golebu2020:
**Admin Tooltip**

- added admin tooltip (with text _Admin Only_') to the `Admin Notes` and `Admin Comment` tabs.
- added notification (with text _This page is for admin only; users do not have access._) to the `Admin Notes` and `Admin Comment` tabs

### DIFF
--- a/app/components/generic/tabs/BasicTabs.tsx
+++ b/app/components/generic/tabs/BasicTabs.tsx
@@ -1,3 +1,4 @@
+import { Tooltip } from '@mantine/core';
 import classNames from 'classnames';
 import _lowerCase from 'lodash-es/lowerCase';
 import _startCase from 'lodash-es/startCase';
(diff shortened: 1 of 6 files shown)


app/components/generic/tabs/BasicTabs.tsx
```diff
@@ -38,20 +39,28 @@ export default function BasicTabs({ tabs, children }: { tabs: ITab[]; children?:
...
+                    : "relative border-transparent text-gray-300 hover:before:content-[''] hover:before:absolute hover:before:w-full hover:before:border-b-3 hover:before:border-gray-300 hover:before:bottom-0 hover:before:left-1/2 hover:before:-translate-x-1/2",
+                );
+
+                const linkingComponent = (
+                  <Link href={tab.href} key={tab.name} className={linkingClassNames}>
+                    {tab.label}
+                  </Link>
+                );
+
+                return tab.name === 'comments' ? (
```
junminahn, line 56: Let's generalize the tab functionality by adding a `tooltip: string` prop to `ITab`. If `tab.tooltip` is not null, the tooltip will display the provided string value.

junminahn (approved):

PR #3658, opened 2024-09-03
fix(3641): resolve a bug displaying multiple app versions in footer

junminahn:


junminahn:
> E2E Tests unsuccessful. Please check.

Thank you for catching it up. E2E will be failed until Artem makes change on his end; let's ignore until then.

PR #3809, opened 2024-09-19
feat(3647): add saving screenshots, tweak e2e tests, sandbox image pulls

funtigr:
feat(3647): add saving screenshots, tweak e2e tests, sandbox image pulls
As part of tweaking tests to run in GitHub actions we faced some obstacles.
First, to understand the Cypress side problems, I added the code to GitHub Action to pass the screenshots form Cypress to GitHub actions Artifacts (thanks Junmin with suggesting it and giving references).
After that I found that the tests in most of the times didn't see the realm provisioned in KeyCloak sandbox instance. I added steps to reach KeyCloak, login and watch for the realm.
Then Junmin found the problem with image pull mechanism and fixed it (fix included in this PR). 
And after that I found the problem with adding Tech Lead to the product. Adding timeout fixed the issue.


junminahn (approved):
👍

PR #3833, opened 2024-09-24
feat(3165): enhance search, filter, and sort on private products

junminahn:
image


junminahn:
> It failed e2e test

yeah. Artem is working on it.

PR #4025, opened 2024-10-18
docs(2692): compare docker image scanning tools

golebu2020:


junminahn (approved):
Overall, it looks really great! No need for updates at this point, but I do have two comments:
1. "Type" and "Cost" seem to give the same information to me.
2. It would be helpful to include the package team's maintenance level, especially for open-source projects.

PR #4171, opened 2024-11-05
chore(3632): rewrite create, edit approval private cloud tests

funtigr:
I have changed the E2E Approval tests for Create and Edit requests in Private cloud, to correspond to new type of steps.


cypress/e2e/create-test.cy.ts
```diff
@@ -4,7 +4,8 @@ export function createRequest(productName: string, POEmail: string, TLEmail: str
   cy.get('input[name="name"]').type(productName);
```
junminahn, line 4: Do we still need this function?

cypress/support/step_definitions/common.ts
```diff
@@ -78,12 +78,16 @@ When(/^User clicks tab "(.*)"$/, (tabText: string) => {
 });
 
 When(/^User clicks and selects "(.*)" in "(.*)"$/, (menuOption: string, menuName: string) => {
-  cy.contains('label', menuName).parents().eq(2).find('input[type="text"]').click();
+  cy.contains('label', menuName).parent().parent().parent().find('input').first().click();
```
junminahn, line 81: it seems not flexible enough. would it be done with `.parents('div')`?

PR #4235, opened 2024-11-13
chore(4176): add auto approval usage metrics for last two weeks

Kolezhanchik:


app/services/k8s/usage-metrics.ts
```diff
@@ -84,28 +127,32 @@ export async function getPodMetrics(
...
-            cpu: normalizeCpu(container.usage.cpu) || 0,
-            memory: normalizeMemory(container.usage.memory) || 0,
+            cpu: normalizeCpu(lastHourUsage.cpu) as number,
+            memory: normalizeMemory(lastHourUsage.memory) as number,
           },
           limits: {
-            cpu: normalizeCpu(resourceDef.limits?.cpu || '0'),
-            memory: normalizeMemory(resourceDef.limits?.memory || '0'),
+            cpu: normalizeCpu(resourceDef.limits?.cpu || 0) as number,
+            memory: normalizeMemory(resourceDef.limits?.memory || 0) as number,
```
junminahn, line 147: let's remove `as number`

app/services/k8s/usage-metrics.ts
```diff
@@ -84,28 +127,32 @@ export async function getPodMetrics(
...
           },
           requests: {
-            cpu: normalizeCpu(resourceDef.requests?.cpu || '0'),
-            memory: normalizeMemory(resourceDef.requests?.memory || '0'),
+            cpu: normalizeCpu(resourceDef.requests?.cpu || 0) as number,
+            memory: normalizeMemory(resourceDef.requests?.memory || 0) as number,
```
junminahn, line 151: let's remove `as number`

app/services/k8s/usage-metrics.ts
```diff
@@ -84,28 +127,32 @@ export async function getPodMetrics(
     // Map over containers to collect their usage, limits, and requests
     const containers = item.containers
       .filter((container) => {
-        if (container.name === 'POD' && container.usage.cpu === '0' && container.usage.memory === '0') return false;
+        if (container.name === 'POD') return false;
         return true;
       })
       .map((container, index) => {
         const resourceDef = podStatus.body.spec?.containers[index]?.resources ?? {};
```
junminahn, line 134: is it reliable to find the container by index, if the container has name, is it better to use the name to fine it? if then, is the name always unique?

Kolezhanchik: When querying pod metrics, Kubernetes adds a pseudo-container named `POD` to represent the pod's management overhead, like network or logging. This is not an actual application container and usually shows zero resource usage `(cpu: '0', memory: '0')`. looks like usually it append as the last array item, but not sure if it is a good idea to relay on `last index`

junminahn (approved):

PR #4477, opened 2024-12-11
Chore/3225 6

Kolezhanchik:


helm/_mailpit/values.yaml
```diff
@@ -1,10 +1,14 @@
 replicaCount: 1
 
+# https://hub.docker.com/r/axllent/mailpit/tags
 image:
   repository: axllent/mailpit
   tag: v1.21.1
   pullPolicy: Always
 
+global:
```
junminahn, line 9: `global` comes from the parent helm chart.

helm/_mailpit/templates/deployment.yaml
```diff
@@ -22,10 +22,10 @@ spec:
         {{- end }}
     spec:
       {{- $imagePullSecrets := concat (.Values.imagePullSecrets | default list) (.Values.global.imagePullSecrets | default list) -}}
-      {{- with $imagePullSecrets }}
-      imagePullSecrets:
-        {{- toYaml . | nindent 8 }}
-      {{- end }}
+        {{- with $imagePullSecrets }}
```
junminahn, line 25: revert this indentation change.

helm/_mailpit/values.yaml
```diff
@@ -73,3 +77,13 @@ nodeSelector: {}
 tolerations: []
 
 affinity: {}
+
+mailpit:
```
junminahn, line 81: 1. this helm chart is for `mailpit`, so having `mailpit` attribute does not makes sense.
2. as mount volume path is hard coded, just hard code the path with `args` as well.

helm/_mailpit/values.yaml
```diff
@@ -73,3 +77,13 @@ nodeSelector: {}
...
+  dbPath: /var/lib/mailpit/mailpit.db
+
+
+persistence:
```
junminahn, line 85: no need to introduce a new attribute, just use `volumes` and revert `volumes` back in deployment template

helm/main/values.yaml
```diff
@@ -201,21 +201,19 @@ weasyprint:
...
   fullnameOverride: pltsvc-mailpit
 
-  resources:
-    limits:
-      cpu: 200m
-      memory: 256Mi
-    requests:
-      cpu: 10m
-      memory: 128Mi
+  dbPath: /var/lib/mailpit/mailpit.db
```
junminahn, line 211: update these two attributes accordingly

PR #4478, opened 2024-12-11
feat(4026): refresh keycloak tokens without signing out

golebu2020:


app/core/auth-options.ts
```diff
@@ -12,7 +14,69 @@ import prisma from '@/core/prisma';
...
+async function processToken(token: Token) {
+  const decodedToken: DecodedToken = jwt.decode(token.accessToken || '') as DecodedToken;
+
+  const newRoles = _get(decodedToken, `resource_access.${AUTH_RESOURCE}.roles`, []);
+  const newSub = decodedToken?.sub ?? '';
+  const newTeams: TeamAccess[] = [];
+
+  _forEach(decodedToken.resource_access ?? {}, (val, key) => {
+    if (key.startsWith(TEAM_SA_PREFIX)) {
+      token.teams.push({ clientId: key, roles: val.roles });
```
junminahn, line 50: you meant `teams.push`?

app/core/auth-options.ts
```diff
@@ -12,7 +14,69 @@ import prisma from '@/core/prisma';
...
+interface DecodedToken {
+  resource_access: Record<string, { roles: string[] }>;
+  sub: string;
+  email: string;
+}
+
+async function processToken(token: Token) {
+  const decodedToken: DecodedToken = jwt.decode(token.accessToken || '') as DecodedToken;
+
+  const newRoles = _get(decodedToken, `resource_access.${AUTH_RESOURCE}.roles`, []);
```
junminahn, line 44: drop `new`, just `roles`.

app/core/auth-options.ts
```diff
@@ -12,7 +14,69 @@ import prisma from '@/core/prisma';
...
+  const newSub = decodedToken?.sub ?? '';
```
junminahn, line 45: drop `new`, just `sub`.

app/core/auth-options.ts
```diff
@@ -12,7 +14,69 @@ import prisma from '@/core/prisma';
...
+  const newTeams: TeamAccess[] = [];
```
junminahn, line 46: drop `new`, just `teams`.

app/core/auth-options.ts
```diff
@@ -12,7 +14,69 @@ import prisma from '@/core/prisma';
...
+  roles: string[];
+}
+
+interface DecodedToken {
+  resource_access: Record<string, { roles: string[] }>;
+  sub: string;
+  email: string;
+}
+
+async function processToken(token: Token) {
```
junminahn, line 41: 1. it seems like `accessToken` is the only attribute required from token? then just `accessToken: string`
2. why this function needs to be `async`?

app/core/auth-options.ts
```diff
@@ -12,7 +14,69 @@ import prisma from '@/core/prisma';
...
+  const decodedToken: DecodedToken = jwt.decode(token.accessToken || '') as DecodedToken;
```
junminahn, line 42: no need to cast twice; 
`const decodedToken = jwt.decode(token.accessToken || '') as DecodedToken;`

app/core/auth-options.ts
```diff
@@ -12,7 +14,69 @@ import prisma from '@/core/prisma';
...
+
+  return {
+    isRefreshTokenExpired: false,
+    roles: newRoles,
+    sub: newSub,
+    teams: newTeams,
+  };
+}
+
+async function getNewTokens(refreshToken: string) {
```
junminahn, line 62: use `try catch` here to check if the call fails; which means the refresh token is expired.

app/core/auth-options.ts
```diff
@@ -233,23 +302,39 @@ export const authOptions: AuthOptions = {
...
-        token.teams = [];
-
-        _forEach(decodedToken.resource_access ?? {}, (val, key) => {
-          if (key.startsWith(TEAM_SA_PREFIX)) {
-            token.teams.push({ clientId: key, roles: val.roles });
-          }
+        Object.assign(token, await processToken(token));
+      }
+
+      try {
```
junminahn, line 314: remove `try catch` here.

app/core/auth-options.ts
```diff
@@ -12,7 +14,69 @@ import prisma from '@/core/prisma';
...
+  const newTeams: TeamAccess[] = [];
+
+  _forEach(decodedToken.resource_access ?? {}, (val, key) => {
+    if (key.startsWith(TEAM_SA_PREFIX)) {
+      token.teams.push({ clientId: key, roles: val.roles });
+    }
+  });
+
+  return {
+    isRefreshTokenExpired: false,
```
junminahn, line 55: I don't think `isRefreshTokenExpired` is required here

app/core/auth-options.ts
```diff
@@ -206,6 +270,11 @@ export const authOptions: AuthOptions = {
       const loweremail = email.toLowerCase();
       const lastSeen = new Date();
 
+      await prisma.user.update({
```
junminahn, line 273: there is `upsert` command just below; why need two database calls that does the same? and this won't work if the user is the new.

app/core/auth-options.ts
```diff
@@ -233,23 +302,39 @@ export const authOptions: AuthOptions = {
...
+          data: {
+            nextTokenRefreshTime: addMinutes(now, USER_TOKEN_REFRESH_MIN),
+          },
         });
+      } catch (error) {}
+
+      if (user) {
+        try {
+          const { access_token } = await getNewTokens(token.refreshToken);
+          token.accessToken = access_token;
```
junminahn, line 329: what about other tokens?
otherwise, how do we update the newest `refresh token` every time?

junminahn (commented):
Please revisit the `jwt` callback to clean up the logic.

app/core/auth-options.ts
```diff
@@ -233,23 +302,39 @@ export const authOptions: AuthOptions = {
...
+            email: token.email,
+            nextTokenRefreshTime: { lt: now },
+          },
+          data: {
+            nextTokenRefreshTime: addMinutes(now, USER_TOKEN_REFRESH_MIN),
+          },
         });
+      } catch (error) {}
+
+      if (user) {
```
junminahn, line 326: remove `try catch` here.

app/core/auth-options.ts
```diff
@@ -12,7 +14,83 @@ import prisma from '@/core/prisma';
...
+  const sub = decodedToken?.sub ?? '';
+  const teams: TeamAccess[] = [];
+
+  _forEach(decodedToken.resource_access ?? {}, (val, key) => {
+    if (key.startsWith(TEAM_SA_PREFIX)) {
+      teams.push({ clientId: key, roles: val.roles });
+    }
+  });
+
+  return {
```
junminahn, line 54:
```js
{
    roles,
    sub,
    teams,
  }
```

app/core/auth-options.ts
```diff
@@ -12,7 +14,83 @@ import prisma from '@/core/prisma';
...
+  });
+  try {
+    const response = await axios.post(tokenEndpoint, params, {
+      headers: {
+        'Content-Type': 'application/x-www-form-urlencoded',
+      },
+    });
+    return response.data;
+  } catch (error) {
+    token.isRefreshTokenExpired = true;
```
junminahn, line 88: just return `null`

app/core/auth-options.ts
```diff
@@ -12,7 +14,83 @@ import prisma from '@/core/prisma';
...
+    Object.assign(token, {
+      accessToken: userAccount.access_token,
+      refreshToken: userAccount.refresh_token,
+      idToken: userAccount.id_token,
+      ...processToken(userAccount.access_token),
+    });
+  }
+}
+
+async function getNewTokens(token: Token) {
```
junminahn, line 72: just `refreshToken: string`

app/core/auth-options.ts
```diff
@@ -233,23 +312,26 @@ export const authOptions: AuthOptions = {
...
-        token.sub = decodedToken?.sub ?? '';
-        token.teams = [];
-
-        _forEach(decodedToken.resource_access ?? {}, (val, key) => {
-          if (key.startsWith(TEAM_SA_PREFIX)) {
-            token.teams.push({ clientId: key, roles: val.roles });
-          }
-        });
+      const now = new Date();
+      updateTokens(account, token);
```
junminahn, line 316: nope..

app/core/auth-options.ts
```diff
@@ -233,23 +312,26 @@ export const authOptions: AuthOptions = {
...
+
+      const { count } = await prisma.user.updateMany({
```
junminahn, line 318: why use `updateMany` all of sudden?

app/services/db/user.ts
```diff
@@ -31,6 +32,7 @@ export async function prepareUserData(user: AppUser, extra = {}) {
     officeLocation: user.officeLocation,
     jobTitle: user.jobTitle,
     upn: user.upn,
+    nextTokenRefreshTime: addMinutes(new Date(), USER_TOKEN_REFRESH_MIN),
```
junminahn, line 35: please leave this file as it was.

junminahn (approved):

PR #4648, opened 2025-01-14
Feat/4506

Kolezhanchik:
image
image


golebu2020:
Could you please provide the ministry associated with the expense authority, as well as the signee details for both the billing page and the events page?

golebu2020 (approved):

PR #4753, opened 2025-01-24
feat(4746): add password input components

junminahn:


app/app/examples/ui/(components)/password-input/page.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const methods = useForm({
+    resolver: zodResolver(validationSchema),
+    defaultValues: {},
+  });
+
+  return (
+    <>
+      <h1 className="font-bold text-2xl mt-4 mb-5">Password Input</h1>
+      <FormProvider {...methods}>
+        <form onSubmit={methods.handleSubmit(console.log)} autoComplete="off">
```
golebu2020, line 29: Is it possible to avoid using console.log in `handleSubmit` and instead pass a proper callback function?

junminahn: @golebu2020 do you mean something like `(formData) => console.log(formData)`?

golebu2020 (approved):

PR #4930, opened 2025-02-12
Chore/4843

Kolezhanchik:


.github/workflows/build-push-maintenance.yml
```diff
@@ -7,6 +7,7 @@ on:
     paths:
     - maintenance/**
     - .github/workflows/build-push-maintenance.yml
+  workflow_dispatch:
```
junminahn, line 10: please remove this

.github/workflows/deploy-maintenance-dispatch.yml
```diff
@@ -0,0 +1,86 @@
...
+        type: choice
+        required: true
+        options:
+        - 'true'
+        - 'false'
+      environment:
+        description: Environment to deploy (dev, test, prod)
+        type: environment
+        required: true
+      imageTag:
```
junminahn, line 17: we don't need this

junminahn (commented):
general workflow steps should be very similar to `prod release` workflow except for the top part, `workflow_dispatch` and `inputs`

junminahn (approved):

PR #5073, opened 2025-02-28
feat(5033): add filters to private cloud analytics page

Kolezhanchik:


app/app/analytics/private-cloud/page.tsx
```diff
@@ -0,0 +1,253 @@
...
+});
+
+const transformMinistryData = (items: { _id: string; value: number }[]) =>
+  items.map(({ _id, value }) => ({ label: ministryKeyToName(_id), value }));
+
+const validClusters = (clusters as string[]).filter((cluster): cluster is Cluster =>
+  Object.values(Cluster).includes(cluster as Cluster),
+);
+
+const mapClusterData = (selectedClusters: Cluster[], ministryData: any[]) => {
```
golebu2020, line 37: We can try making this more elegant using `reduce` (for instance): 

```
const mapClusterData = (selectedClusters: Cluster[], ministryData: any[]) => {
  const clusterDataMapping = selectedClusters.reduce((acc, cluster, index) => {
    const clusterIndex = validClusters.indexOf(cluster);
    acc[cluster] = clusterIndex !== -1 && ministryData[clusterIndex + 1] ? ministryData[clusterIndex + 1] : [];
    return acc;
  }, {} as Record<Cluster, { _id: string; value: number }[]>);

  return selectedClusters.reduce((acc, cluster) => {
    acc[cluster] = transformMinistryData(clusterDataMapping[cluster] || []);
    return acc;
  }, {} as Record<Cluster, ReturnType<typeof transformMinistryData>>);
};
```

app/app/analytics/private-cloud/page.tsx
```diff
@@ -0,0 +1,253 @@
...
+
+  const pieChartData: Record<string, { label: string; value: number }[]> = {
+    All: allClusterData,
+    ...mappedClusterData,
+  };
+
+  return (
+    <div>
+      <h1 className="text-xl lg:text-2xl 2xl:text-4xl font-semibold leading-7 text-gray-900">
+        Private Cloud Data Analytics
```
golebu2020, line 99: Add some bottom padding here

golebu2020:
When I export `users with quote requests', the `last seen columns` is an exponential number. Please fix that.

golebu2020:
Similarly, when I export the data for `Requests over time`, the date does not include the year.

golebu2020:
The buttons' margin bottoms are not consistent across all graphs, please make them consistent.

golebu2020:
Is it possible to embed the export button within the real estate of the graph, probably at the bottom within the graph container. The button appears to be floating outside which could lead to a false judgement for users.

golebu2020:
For the user picker, we could probable replace `Click to select member` with `Click to select user`. I think the previous was used to describe additional members.

golebu2020:
The container for `Users with quota edit requests` also needs some shadow for consistency sake.

golebu2020:
You can also remove the tag 'Date Range` above the select button and instead let the text on the button be `Select date range`

Kolezhanchik:
> The container for `Users with quota edit requests` also needs some shadow for consistency sake.

this is a common component and changing text means to change it in all places component is used

Kolezhanchik:
> You can also remove the tag 'Date Range`above the select button and instead let the text on the button be`Select date range`

this is a common component and changing text means to change it in all places component is used

Kolezhanchik:
> The container for `Users with quota edit requests` also needs some shadow for consistency sake.

it replicates the original page layout, and it is out of ticket scope, you can create ticket to revise analytics page layout design

Kolezhanchik:
> You can also remove the tag 'Date Range`above the select button and instead let the text on the button be`Select date range`

this is a common component and changing text means to change it in all places component is used

Kolezhanchik:
> it replicates the original page layout, and it is out of ticket scope, you can create ticket to revise analytics page layout design

it replicates the original page layout, and it is out of ticket scope, you can create ticket to revise analytics page layout design

Kolezhanchik:
> Similarly, when I export the data for `Requests over time`, the date does not include the year.

it replicates the original download data, and it is out of ticket scope, you can create ticket to revise analytics page download data format

junminahn (commented):
1. naming conventions
 - get-active-products > active-products
 - privateCloudAnalytics -> getPrivateCloudAnalyticData
 - React component function use `function`
3. split download API endpoints per sections
4. `/api/analytics/private-cloud/filter` -> `/api/analytics/private-cloud/chart-data`
5. one component per file under analystics frontend page
6. drop `return` in `return await downloadPrivateCloudAnalytics(exportParams);`
7. Loader on the page level
8. on the page loader, either
  - have the sibling content OR
  - just use `Loader`
9. move `Filter` component into a separate component file in the same directory
10. create each child components and move them into a single file in the same directory

PR #5103, opened 2025-03-04
feat(4982): add private cloud cluster badge

junminahn:


app/services/api-test/public-cloud/products.ts
```diff
@@ -54,8 +54,12 @@ export async function getPublicCloudProject(licencePlate: string) {
   return result;
 }
 
-export async function editPublicCloudProject(licencePlate: string, data: PublicCloudEditRequestBody) {
+export async function editPublicCloudProject(licencePlate: string, data: any) {
```
Kolezhanchik, line 57: why it is better to use `any` instead of `PublicCloudEditRequestBody`?

Kolezhanchik (approved):

PR #5150, opened 2025-03-10
Feat/5080

Kolezhanchik:
image
image


app/components/users/UserProfile.tsx
```diff
@@ -12,10 +12,11 @@ export type UserPickerData = Pick<User, 'email' | 'firstName' | 'lastName' | 'mi
 interface Props {
   data?: UserPickerData;
   onClick?: () => void;
+  noUserSelectedText?: string;
```
junminahn, line 15: `noUserSelectedText` -> `text`

junminahn (commented):
let's move the tooltip into `UserProfile` component level and no need to display `delete user`

junminahn (approved):

PR #5153, opened 2025-03-11
feat(4927): add reusable TooltipTableHeader component

Kolezhanchik:


junminahn (commented):
It is not the last sign-in date; it is the datetime when the user was last active in the registry app. so let's say.
`The last active date/time in the registry`

junminahn (approved):

PR #5173, opened 2025-03-13
chore(5059): implement auto-versioning with strict tag validation

Kolezhanchik:


junminahn (commented):
can you add following in the line 45?
```
NEXT_VERSION=${NEXT_VERSION//v}
```

junminahn (approved):

PR #5210, opened 2025-03-14
feat(4750): send cancellation notification emails

junminahn:


golebu2020:
It is possible to include the request type in the body of the email template?

golebu2020 (approved):

PR #5252, opened 2025-03-20
feat(4743): show users with incomplete idir profile

golebu2020:
Screenshot 2025-03-25 at 2 56 10 PM


app/app/api/msgraph/search/route.ts
```diff
@@ -1,16 +1,30 @@
...
 import { OkResponse } from '@/core/responses';
 import { prepareUserData } from '@/services/db';
 import { listUsersByEmail } from '@/services/msgraph';
+import { AppUser } from '@/types/user';
 
 const userSearchBodySchema = z.object({
   email: z.string().max(40),
 });
 
+function syncUsers(appUsers: AppUser[], dbUsers: any) {
```
Kolezhanchik, line 18: dbUsers has type `User[]`, but also it is possible to `pick` only `email` field of `User`

app/app/api/msgraph/search/route.ts
```diff
@@ -1,16 +1,30 @@
...
+  const mappedbUsers = keyBy(dbUsers, 'email');
+
```
Kolezhanchik, line 20: maybe something like `dbUsersByEmail` to make the name more informative?

app/app/api/msgraph/search/route.ts
```diff
@@ -1,16 +1,30 @@
...
+  forEach(appUsers, (appUser, index) => {
+    if (appUser && appUser.email && mappedbUsers[appUser.email]) {
+      appUsers[index] = mappedbUsers[appUser.email];
```
Kolezhanchik, line 23: is this condition needed here `appUser && `? you will check `user` with email on next step

golebu2020: appUser checks that we're not dealing with a `undefined` situation.

Kolezhanchik: yes, but `appUser.email` also checks user and email

app/app/api/msgraph/search/route.ts
golebu2020: We can't just pick only the email because if we do, then we'll have a user pick that displays only the `email`. Type of `User[]` cannot be used because some fields is missing from the user created in the database (for instance, `displayName`.

app/app/api/msgraph/search/route.ts
golebu2020: okay

app/app/api/msgraph/search/route.ts
Kolezhanchik: ok, but it is still possible to use same type `AppUser[]` or `User` instead of `any`

app/app/api/msgraph/search/route.ts
Kolezhanchik: it wasn't about function, it was about variable `mappedbUsers`

junminahn (commented):
Please visit this page to update your missing profile information:
<ExternalLink href="https://www2.gov.bc.ca/gov/content/governments/services-for-government/information-management-technology/id-services">IDIR Services - Government of BC</ExternalLink>

junminahn (commented):
1. removed the function sync... is a good choice, but please review the return value of the invalid user data; I would suggest define the return data type as the same as the matching frontend's service endpoint's data type.
2. please introduce a way to test the invalid user in the local environment in permanent way.

junminahn (approved):

PR #5281, opened 2025-03-25
Chore/5041 1

Kolezhanchik:


app/app/api/security-config/route.ts
```diff
@@ -21,8 +21,8 @@ export const PUT = apiHandler(async ({ body, session }) => {
   if (count === 0) {
     count =
       body.context === ProjectContext.PRIVATE
-        ? await prisma.privateCloudRequestedProject.count(existQuery)
-        : await prisma.publicCloudRequestedProject.count(existQuery);
+        ? await prisma.publicCloudRequestData.count(existQuery)
```
junminahn, line 24: here is a typo

data-migrations/migrations/20250320210834-rename_project_to_product.js
```diff
@@ -0,0 +1,29 @@
+export const up = async (db, client) => {
+  // Rename collection: PrivateCloudProject → PrivateCloudProduct
+  const oldPrivateCloudProject = await db.listCollections({ name: 'PrivateCloudProject' }).toArray();
```
junminahn, line 3: let's pull all collection name and use `some` method to find target collections
```js
const collections = await db.listCollections({}, { nameOnly: true }).toArray();
```

junminahn (commented):
let's revert all changes on the old `data migration scripts` as they are historical migration scripts and won't be run again in live environments.

junminahn (approved):

PR #5325, opened 2025-03-28
fix(5304): restrict same id users to be added to team contacts

Kolezhanchik:


app/components/modal/userPicker.tsx
```diff
@@ -32,6 +33,11 @@ function WarningMessage({ message }) {
   );
 }
 
+const isUserDuplicate = (users: ({ id?: string | null } | null | undefined)[], userId: string | undefined | null) => {
+  if (!userId) return false;
+  return users.filter((u) => u?.id === userId).length > 1;
```
golebu2020, line 38: Can you use this instead?: 
``` Javascript
 return !!userId && users.filter(user => user?.id === userId).length > 1;
```

app/components/shared/TeamContacts.tsx
```diff
@@ -37,7 +37,10 @@ export default function TeamContacts({ disabled, userAttributes }: Props) {
     const handleUserChange = async () => {
       if (disabled) return;
 
-      const { state } = await openUserPickerModal({ initialValue: user }, { initialState: { user } });
+      const initialState = role === 'Expense Authority (EA)' ? { user } : { user, users };
```
golebu2020, line 40: I believe this can be made better

app/components/modal/userPicker.tsx
Kolezhanchik: why would I check next condition if it is possible to return false if `!userId`?

junminahn (approved):

PR #5336, opened 2025-04-01
feat(4721): resolve test failures for new provision endpoint

golebu2020:


app/helpers/mock-resources.ts
```diff
@@ -114,3 +118,36 @@ export function createSamplePrivateCloudCommentData(args?: { data?: Partial<Pris
...
+    },
+    secret,
+  });
+  return {
+    Authorization: 'Bearer ' + signature,
+    'Content-Type': 'application/json',
+  };
+}
+
+export function getProvisionTestAuthHeader() {
```
junminahn, line 139: let's convert it to `getTeamServiceAccountAuthHeader` to support wider usage.

app/helpers/mock-resources.ts
```diff
@@ -114,3 +118,36 @@ export function createSamplePrivateCloudCommentData(args?: { data?: Partial<Pris
 
   return _data;
 }
+
+export function getUserTestAuthHeader(user?: AppUserWithRoles) {
```
junminahn, line 122: let's rename it to `getUserServiceAccountAuthHeader`

app/helpers/mock-resources.ts
```diff
@@ -114,3 +118,36 @@ export function createSamplePrivateCloudCommentData(args?: { data?: Partial<Pris
...
+    Authorization: 'Bearer ' + signature,
+    'Content-Type': 'application/json',
+  };
+}
+
+export function getTeamServiceAccountAuthHeader() {
+  const signature = jws.sign({
+    header: { alg: 'HS256', typ: 'JWT' },
+    payload: {
+      roles: 'private-admin,public-admin',
```
junminahn, line 143: when convert the method more generic, then the logic should support it as well.. does all service account have the same roles?
- also please create a function `getServiceAccountAuthHeader` that encapsulates the common logic from both `getUserServiceAccountAuthHeader` and `getTeamServiceAccountAuthHeader`

app/services/api-test/core.ts
```diff
@@ -131,7 +131,14 @@ export async function mockSessionByRole(role?: string) {
     if (!user) mockedGetServerSession.mockResolvedValue(null);
     else {
       const mockSession = await generateTestSession(user.email);
-      mockedGetServerSession.mockResolvedValue(mockSession);
+      const rolesToAdd = ['private-admin', 'public-admin', 'service-account'];
```
junminahn, line 134: why do we need this additional block?

junminahn (commented):
In general, do we still need these two method?
- mockUserServiceAccountByEmail
- mockUserServiceAccountByRole

app/services/api-test/core.ts
golebu2020: We'll encounter authorization errors without including those roles.

golebu2020:
Yes, we still need the methods: `mockUserServiceAccountByEmail` and 
`mockUserServiceAccountByRole`

junminahn (approved):

PR #5377, opened 2025-04-03
refactor(5235): use k8s package to check deletion availability

junminahn:


app/services/k8s/reads/deletion-check.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  const result = {
+    namespace: false,
+    pods: false,
+    pvc: false,
+    artifactory: false,
+  };
+
+  const namespaceNames = ['dev', 'test', 'dev', 'tools'].map((env) => `${licencePlate}-${env.toLowerCase()}`);
```
Kolezhanchik, line 30: is it possible to use here ? 

```
import  {ProjectSetNamespace} from "/app/helpers/openshift.ts"

const namespaces: string[] = Object.values(ProjectSetNamespace);

const namespaceNames = Object.values(ProjectSetNamespace).map(
  (env) => `${licencePlate}-${env.toLowerCase()}`
);
```

Kolezhanchik (approved):

Kolezhanchik (approved):

PR #5385, opened 2025-04-08
Feat/5124

golebu2020:


helm/tools/dags/_msgraph.py
```diff
@@ -0,0 +1,36 @@
...
+        response.raise_for_status()
+        return response.json()
+
+    def _get_access_token(self):
+        token_data = self._request_token()
+        return token_data.get("access_token", "")
+
+    def get_azure_users(self):
+        access_token = self._get_access_token()
+        graph_url = "https://graph.microsoft.com/v1.0/users"
```
Kolezhanchik, line 29: /users endpoint is paginated by default (probably 100) , is it possible to use here `@odata.nextLink` and fetch all pages.

helm/tools/dags/sync_db_with_azure_ad_prod.py
```diff
@@ -0,0 +1,31 @@
...
+    start_date=datetime.now() - timedelta(days=1),
+    catchup=False,
+) as dag:
+    t1 = PythonOperator(
+        task_id="sync-db-users-with-azure-ad-prod",
+        python_callable=sync_db_users_with_azure_ad,
+        op_kwargs={
+            "mongo_conn_id": MONGO_CONN_ID,
+            "ms_graph_api_tenant_id": PROD_MS_GRAPH_API_TENANT_ID,
+            "ms_graph_api_client_id": PROD_MS_GRAPH_API_TENANT_ID,
```
Kolezhanchik, line 25: `ms_graph_api_tenant_id` and `ms_graph_api_client_id` same value here

junminahn (commented):
1. What about the `live dev` environment?  
2. For each live environment script:
   1. Configure `<env>_MS_GRAPH_API_TENANT_ID` directly, instead of using `<env>_MS_GRAPH_API_AUTHORITY`, as the latter is unused.
   2. Remove environment-specific prefixes when retrieving environment variables.
3. In the main logic file:
   1. Ensure it only fetches `active` users from the database.
   2. Fetch the corresponding MS user per DB user instead of listing all MS users; THINK the reasoning behind this.
   3. Review the `processMsUser` function to confirm which attributes need to be processed and updated in the DB.
   4. Confirm how to retrieve all required attributes from MS Graph API.

helm/tools/dags/_sync_azure_ad_with_db.py
```diff
@@ -0,0 +1,49 @@
...
+                    "providerUserId": azure_user.get("id"),
+                    "ministry": parse_ministry_from_display_name(
+                        azure_user.get("displayName"),
+                    ),
+                    "firstName": azure_user.get("givenName"),
+                    "lastName": azure_user.get("surname"),
+                    "email": azure_user.get("mail").lower(),
+                }
+
+                update_data = {k: v for k, v in update_data.items() if v is not None}
```
junminahn, line 42: we need to set empty/default values for each attributes if not exist, does individual `get` method takes the second argument to set the default values above?

helm/tools/dags/_msgraph.py
```diff
@@ -0,0 +1,48 @@
...
+
+        try:
+            response = requests.get(graph_url, headers=headers)
+            if response.status_code == 401 and retry:
+                self.access_token = self._get_access_token()
+                return self.fetch_azure_user(matching_email, retry=False)
+
+            response.raise_for_status()
+            users = response.json().get("value", [])
+            return users[0] if users else None
```
junminahn, line 45: let's make sure the status code is `200`

junminahn (approved):

PR #5419, opened 2025-04-10
feat(5343): redirect anauthorized user to original url after login

Kolezhanchik:


app/core/client-page.tsx
```diff
@@ -46,13 +46,24 @@ function createClientPage<TPathParams extends ZodType<any, any>, TQueryParams ex
...
         updateSession();
+        if (status === 'authenticated' && localStorage.getItem('postLoginRedirect')) {
+          localStorage.removeItem('postLoginRedirect');
+        }
       }, []);
 
+      function handleAccessRedirect() {
+        if (status === 'unauthenticated') {
+          localStorage.setItem('postLoginRedirect', pathname);
+          return null;
```
junminahn, line 62: it shouldn't return here as it still needs to be redirected.

app/core/client-page.tsx
```diff
@@ -46,13 +46,24 @@ function createClientPage<TPathParams extends ZodType<any, any>, TQueryParams ex
...
+        }
+        return router.push(fallbackUrl);
```
junminahn, line 64: remove `return` here as there is no effect.

app/core/client-page.tsx
```diff
@@ -65,15 +76,15 @@ function createClientPage<TPathParams extends ZodType<any, any>, TQueryParams ex
       if (roles && roles.length > 0) {
         const allowed = arrayIntersection(roles, _roles).length > 0;
         if (!allowed) {
-          return router.push(fallbackUrl);
+          handleAccessRedirect();
```
junminahn, line 79: after handle redirect; please `return` in the next line to ensure no any logic keeps running after; and same for other places.

junminahn (approved):

PR #5422, opened 2025-04-10
feat(5382): implement backend support for Private Cloud unit price page

junminahn:


Kolezhanchik (approved):

app/core/auth-options.ts
funtigr, line 1: @junminahn  
As far as I see here, `viewPrivateCloudUnitPrices` permission is given to `isPublicAdmin`, `isPublicEditor`, `isPublicReader`. 
Shouldn't they be `Private`?

PR #5502, opened 2025-04-24
feat(5473): make expenseAuthorityId field mandatory

golebu2020:


data-migrations/migrations/20250423230707-replace_null_eaid_with_poid.js
```diff
@@ -0,0 +1,23 @@
+export const up = async (db, client) => {
+  const collectionNames = ['PublicCloudProduct', 'PublicCloudRequestData'];
+  await Promise.all(
+    collectionNames.map(async (collectionName) => {
+      const collection = await db.collection(collectionName);
```
junminahn, line 5: `collection` method is not async

junminahn (commented):
also please use `Transaction` in the data migration script;
also please make note how many projects are affected in the public cloud projects & request data

junminahn (approved):

PR #5735, opened 2025-05-29
fix(5714): make cost charts responsive

Kolezhanchik:


Kolezhanchik:
> When I try to render it locally the chart looks miniaturized. Can we increase the height of the container? Screenshot 2025-05-29 at 11 39 45 AM

image

PR #5929, opened 2025-06-24
feat(5856): replace team api account table with cc table

golebu2020:
Screenshot 2025-06-24 at 10 45 19 AM

### DIFF
--- a/app/app/team-api-accounts/page.tsx
+++ b/app/app/team-api-accounts/page.tsx
@@ -123,7 +123,7 @@ export default TeamApiAccountsPage(({ session }) => {
       </Table.Tr>
     );
   } else {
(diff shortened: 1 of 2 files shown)


app/app/team-api-accounts/page.tsx
```diff
@@ -41,59 +123,17 @@ export default TeamApiAccountsPage(({ session }) => {
       </Table.Tr>
     );
   } else {
-    rows = (apiAccounts || []).map((account) => {
+    (apiAccounts || []).map((account) => {
```
Kolezhanchik, line 126: with new code `.map()` doesn't return anything, `.forEach` seems a better option

Kolezhanchik (approved):

PR #5980, opened 2025-07-03
refactor(5845): update MS graph API services using Axios configs

junminahn:


app/services/msgraph/axios.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (graphAPIProxy) {
+      // If using a proxy, ensure the axios instance uses the proxy agent
+      config.httpAgent = graphAPIProxy;
+      config.httpsAgent = graphAPIProxy;
+    }
+
+    return config;
+  },
+  (error) => {
+    return Promise.reject(error);
```
Kolezhanchik, line 57: error here might be anything `string`, `object` right? `Promise.reject` expects `error`

app/services/msgraph/index.ts
```diff
@@ -50,30 +38,48 @@ const userAttributes = [
...
+      },
+    })
+    .then((res) => res.data)
+    .catch((err) => {
+      if (err.status === 404) {
+        return null; // User not found
+      }
+
+      logger.error('Error fetching user by UPN:', err.message ?? String(err));
+      return null;
```
Kolezhanchik, line 58: it returns `null` in all cases `err.status === 404` and all others

app/services/msgraph/index.ts
```diff
@@ -84,9 +90,19 @@ export async function getUserByEmail(email: string) {
...
+  const result = await instance
+    .get<ArrayBuffer>(`/users/${upn.toLowerCase()}/photo/$value`, { responseType: 'arraybuffer' })
+    .then((res) => res.data)
+    .catch((err) => {
+      if (err.status === 404) {
+        return null; // User photo not found
+      }
+
+      logger.error('Error fetching user photo:', err.message ?? String(err));
+      return null;
```
Kolezhanchik, line 104: it returns `null` in all cases `err.status === 404` and all others

app/services/msgraph/axios.ts
junminahn: `Promise.reject` expect anything actually; and if there is type issue, the build would throw an exception

app/services/msgraph/index.ts
junminahn: in Axios, 4xx treated as an error, so need to handle separately

app/services/msgraph/index.ts
junminahn: the same as above

Kolezhanchik (approved):

PR #6142, opened 2025-07-23
feat(5915): rename createdByEmail and decisionMakerEmail to createdBy…

Kolezhanchik:
…Id and decisionMakerId


app/components/table/TableBodyPrivateProducts.tsx
```diff
@@ -92,7 +92,7 @@ export default function TableBodyPrivateProducts({ rows, isLoading = false }: Ta
             <div className="md:col-span-2 lg:col-span-3">
               {row.activeRequest && (
                 <PrivateCloudActiveRequestBox
-                  request={{ ...row.activeRequest, licencePlate: row.licencePlate }}
+                  request={{ ...row.activeRequest, licencePlate: row.licencePlate, createdBy: row.createdBy }}
```
junminahn, line 95: why do we want to use a product's created user information rather than the request itself?

app/components/table/TableBodyPublicProducts.tsx
```diff
@@ -85,7 +85,9 @@ export default function TableBodyPublicProducts({ rows, isLoading = false }: Tab
 
             <div className="md:col-span-2 lg:col-span-3">
               {row.activeRequest && (
-                <PublicCloudActiveRequestBox request={{ ...row.activeRequest, licencePlate: row.licencePlate }} />
+                <PublicCloudActiveRequestBox
+                  request={{ ...row.activeRequest, licencePlate: row.licencePlate, createdBy: row.createdBy }}
```
junminahn, line 89: the same as above

data-migrations/migrations/20250723175849-rename_and_populate_createdbyemail_to_createdbyid.js
```diff
@@ -0,0 +1,40 @@
+export const up = async (db, client) => {
```
junminahn, line 1: 1. we do not need to rename the fields.
2. let's fetch all users at once to reduce the DB query calls.

junminahn (approved):

PR #6290, opened 2025-08-12
chore(6088): display role change if user has more than one role

Kolezhanchik:
image

### DIFF
--- a/app/services/db/members-history.ts
+++ b/app/services/db/members-history.ts
@@ -105,7 +105,7 @@ export async function buildMembersHistory(requests: RequestWithData[]) {
 
       if (filteredNewRoles.length > 0) {
         newRoles = filteredNewRoles;
-      } else if (isNewUser || filteredNewRoles.length === 0) {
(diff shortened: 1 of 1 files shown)


app/services/db/members-history.ts
```diff
@@ -105,7 +105,7 @@ export async function buildMembersHistory(requests: RequestWithData[]) {
 
       if (filteredNewRoles.length > 0) {
         newRoles = filteredNewRoles;
-      } else if (isNewUser || filteredNewRoles.length === 0) {
+      } else if ((isNewUser || filteredNewRoles.length === 0) && request.changes?.parentPaths.includes('members')) {
```
MonicaG, line 108: I think this is fine, but I don't have enough context to know for sure.

For my own understanding would you mind explaining: 

- where does the `parentPaths` comes from? I'm assuming that `changes` is a field from the db that contains some kind of structure that includes `parentPaths`? Is that correct?
- is `members` a common value used in various places? I'm wondering about using a constant vs a string to avoid typo mistakes
- will `members` always be lower case? Does the code need to consider the string's case?

MonicaG (approved):

PR #6292, opened 2025-08-13
feat(6087): restrict submitting Additional Team Members without roles

Kolezhanchik:


MonicaG (approved):
Looks good. I don't have all the context, but from what  understand, it looks good to me.

PR #6407, opened 2025-08-29
chore(sandbox): updated config for running sandbox on MacOS and updat…

MonicaG:
…ed related documentation

This PR:

* Updates the `docker-compose-arm64.yml` to work with MacOS
* Updated documentation regarding how to run the site locally
* Removed instructions on Docker Desktop as it is no longer allowed


junminahn (approved):
👍

PR #7478, opened 2026-05-26
doc(2804): add comprehensive Registry roles and permissions guide

Kolezhanchik:


docs/business-logic/registry-roles-guide.md
```diff
@@ -0,0 +1,664 @@
...
+
+### Can
+
+-   View Private Cloud analytics dashboards
+
+### Cannot
+
+-   Edit products
+-   Review requests
+
```
MonicaG, line 248: Does this section need a "Can access" section? Can private-analyser access products not assigned to the user?

docs/business-logic/registry-roles-guide.md
```diff
@@ -0,0 +1,664 @@
...
+### Can
+
+-   View Public Cloud analytics dashboards
+
+### Typical Users
+
+-   Reporting teams
+-   Analytics users
+
```
MonicaG, line 387: Does this section need a "Can access" section? Can public-analyser access products not assigned to the user?

docs/business-logic/registry-roles-guide.md
```diff
@@ -0,0 +1,664 @@
...
+---
+
+# Billing and Finance Roles
+
```
MonicaG, line 391: Do the roles below need a "Can access" section too?

docs/business-logic/registry-roles-guide.md
```diff
@@ -0,0 +1,664 @@
...
+## `billing-reviewer`
+
+### Purpose
+
+Billing review role.
+
+### Can
+
+-   Review Public Cloud billing
+-   View Public Cloud billing
```
MonicaG, line 403: The roles below have access to both private and public billing. The docs here for the `billing-reviewer` state they only have access to public cloud billing. Is that correct? Or do they have access to private cloud billing too?

MonicaG (commented):
Looks good. I left a few comments.

- Some of the roles had a "Can access" section but others did not. I was wondering if it would make sense to have the "Can access" section for all the roles?
- I wasn't sure if the billing-reviewer has access to both public and private cloud or just  the public cloud.

docs/business-logic/registry-roles-guide.md
Kolezhanchik: this role has same `can access` as ` can` so, it would be duplication, this role alone doesn't give any permissions except `View Private Cloud analytics dashboards`

docs/business-logic/registry-roles-guide.md
Kolezhanchik: this role has same `can access` as ` can` so, it would be duplication, this role alone doesn't give any permissions except `View Public Cloud analytics dashboards`

docs/business-logic/registry-roles-guide.md
Kolezhanchik: this role has same `can access` as ` can` so, it would be duplication, this role alone doesn't give any permissions except `Review Public Cloud billing` and `View Public Cloud billing`

docs/business-logic/registry-roles-guide.md
Kolezhanchik: for now private cloud billing doesn't exist, so it would be wrong to mislead guide reader saying that private cloud billing is available

MonicaG (approved):

PR #7480, opened 2026-06-02
docs(2830): replace RC with MS Teams in docs

Kolezhanchik:


.github/ISSUE_TEMPLATE/onboarding-a-new-team-member----name-.md
```diff
@@ -25,7 +25,7 @@ Please use this epic to create various tickets for onboarding a new team member
 -   [ ] Send the meeting invitations for all the scrum and team meeting series
 -   [ ] Add a new team member to the GitHub orgs and repos
 -   [ ] Add a new team member to the ZenHub board
```
MonicaG: I know this ticket was just about updating RC -> MS references, but do we even need this `ISSUE_TEMPLATE` anymore? 

This seems to be about setting up someone new in the DevX team, which should be handled outside of this.

If so, we could just get rid of this altogether, which also solves the RC->MS updates

docs/development-setup/onboarding-offboarding.md
MonicaG: Do we need these docs here? Can we just delete them.

Most of this information is specific to onboarding to the DevX team, which is probably better handled in confluence, then in this app.

I think the docs should be specific to this app, and not general team onboarding info.

docs/business-logic/registry-roles-guide.md
MonicaG, line 1: Were the changes on this file meant to be part of the PR? They don't seem to be related to RC->MS changes.

It's ok if they aren't and you're wanting to bundle docs updates into one update. 

But just double checking this was intended.

If intended, maybe update the title and/or description to reflect these changes were included as well

Kolezhanchik: yes, that was intended and i mentioned it on stand up, i created roles guide based on original request for networking options, but it was changed

.github/ISSUE_TEMPLATE/onboarding-a-new-team-member----name-.md
Kolezhanchik: agree

Pilargit12: Agree, I think this was also specific to tasks in private cloud as they had a Google shared calendar and zenhub.

docs/development-setup/onboarding-offboarding.md
Pilargit12: Agree, thank you for removing this.

Pilargit12 (commented):
Just need to look at the changes requested, resend for review when ready. Thank you Zhanna

Pilargit12 (approved):
Looks good!

PR #7508, opened 2026-06-12
docs(devxt-2909): updated developer docs and associated dev scripts

MonicaG:
This PR does the following:

* Update the developer docs where commands were failing
* Added instructions on how to get test username/passwords for development
* Updated docs to use the Makefile `sandbox` or `localmac` targets rather than manually setting up docker compose
* Modified Makefile `sandbox` and `localmac` to take a `SBD` flag to optionally run in detached mode
* Updated copy-db script
  * removed asdf tool install to the top level .tool-versions file
  * added error checking of file download
    * Note: The error checking added a bit of code to the copy-db script and I'm totally fine reverting it back to what is was originally

### DIFF
--- a/.bin/copy-db.sh
+++ b/.bin/copy-db.sh
@@ -16,8 +16,8 @@ mkdir -p "$tmp_dir"
 
 pod_name=$(oc get pods | grep pltsvc-db-backup- | awk '{print $1}')
 
-if [ -z "$pod_name" ]; then
-    echo "No pod found matching 'pltsvc-db-backup-'"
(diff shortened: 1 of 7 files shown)


Kolezhanchik:
PR looks good overall. The documentation updates are valid and the move toward using the Makefile targets (sandbox / localmac) simplifies local setup. The copy-db changes also add some useful validation around downloads.

The only issue I noticed is that the pre-commit checks are currently failing. Once those failures are addressed and the checks are passing, this should be good to go from my side.

Kolezhanchik (approved):

PR #7563, opened 2026-06-19
chore(2831): strengthen HTTP security headers

Kolezhanchik:


app/middleware.ts
```diff
@@ -29,12 +29,12 @@ export function middleware(request: NextRequest) {
...
     // `style-src 'self' 'nonce-${nonce}'`,
-    `script-src 'self' 'unsafe-inline'`,
-    `style-src 'self' 'unsafe-inline'`,
+    `script-src 'self' 'nonce-${nonce}'`,
+    `style-src 'self' 'nonce-${nonce}'`,
     `img-src 'self' blob: data: ${gravatar_com}`,
     `connect-src 'self' ${gravatar_com} ${loginproxy_gov}`,
     `frame-src ${loginproxy_gov}`,
-    `frame-ancestors ${loginproxy_gov}`,
+    `frame-ancestors 'none'`,
```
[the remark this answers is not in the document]
Kolezhanchik: `X-Frame-Options` was set to `DENY`

app/middleware.ts
```diff
@@ -34,7 +34,7 @@ export function middleware(request: NextRequest) {
...
```
danilomeireles, line 37: One thing I noticed: the PR title mentions strengthening the security headers, but the `unsafe-inline` on `script-src/style-src` (lines 32–33) is still there, and the `nonce-based` version is commented out just above it (lines 30–31). That inline-script/style allowance is really the main XSS exposure, so I'd love to see it tackled too. Or if it's out of scope for this PR, maybe we just tweak the title/description so it's clear that's still pending? Either way works for me, just want to make sure it doesn't get forgotten.

Kolezhanchik: I initially looked into switching to a nonce-based CSP, but during the review it became clear that the application currently does not propagate the nonce to Next.js runtime scripts and styles. Enabling a nonce-based policy in this PR could potentially break hydration and runtime behavior.

So I decided to limit the scope of this PR to lower-risk header hardening changes:

* `frame-ancestors 'none'`
* `X-Frame-Options: DENY`
* existing HSTS and `nosniff` protections

The `unsafe-inline` allowances are still in place for now and should be addressed in a follow-up change once nonce propagation is properly implemented across the application.

I’ll update the PR description to reflect this more clearly.

PR #7697, opened 2026-07-15
Revert "chore(deps): update dependency @keycloak/keycloak-admin-client to v26.7.0"

Kolezhanchik:
Reverts bcgov/platform-services-registry#7673


Kolezhanchik:
Reverting @keycloak/keycloak-admin-client to 26.6.4. The upgrade to 26.7.0 introduced a regression causing all endpoints that initialize the Keycloak admin client to fail during kcAdminClient.auth() with TypeError: Cannot read properties of undefined (reading 'split'). Rolling back restores the previous working behavior until an upstream fix is available.

PR #7831, opened 2026-08-05
chore(2988): add repository availability checkbox and remove organiza…

Kolezhanchik:
…tion restrictions

### DIFF
--- a/app/app/private-cloud/products/(product)/[licencePlate]/edit/page.tsx
+++ b/app/app/private-cloud/products/(product)/[licencePlate]/edit/page.tsx
@@ -98,13 +98,14 @@ export default privateCloudProductEdit(({ session }) => {
 
   useEffect(() => {
     if (!snap.currentProduct) return;
-
+    const repositories = snap.currentProduct.repositories ?? [];
     setDisabled(!snap.currentProduct?._permissions.edit);
 
     methods.reset(
       {
         ...snap.currentProduct,
-        repositories: snap.currentProduct.repositories ?? [],
+        hasRepositories: snap.currentProduct.hasRepositories ?? (repositories.length > 0 ? true : null),
+        repositories,
         isAgMinistry: false,
         isAgMinistryChecked: true,
       },
@@ -116,7 +117,9 @@ export default privateCloudProductEdit(({ session }) => {
 
   const isSubmitEnabled = Object.keys(formState.dirtyFields).length > 0;
 
-  if (!snap.currentProduct) {
+  const currentProduct = snap.currentProduct;
+
+  if (!currentProduct) {
     return null;
   }
 
@@ -130,7 +133,7 @@ export default privateCloudProductEdit(({ session }) => {
         disabled: isDisabled,
         clusterDisabled: true,
         mode: 'edit',
-        canToggleTemporary: snap.currentProduct._permissions.toggleTemporary,
+        canToggleTemporary: currentProduct._permissions.toggleTemporary,
       },
     },
     {
@@ -140,7 +143,7 @@ export default privateCloudProductEdit(({ session }) => {
       Component: TeamContacts,
       componentArgs: {
         isTeamContactsDisabled: isDisabled,
-        isAdditionalMembersDisabled: isDisabled || !snap.currentProduct._permissions.manageMembers,
+        isAdditionalMembersDisabled: isDisabled || !currentProduct._permissions.manageMembers,
       },
     },
     {
@@ -176,7 +179,7 @@ export default privateCloudProductEdit(({ session }) => {
           onSubmit={methods.handleSubmit(async (formData) => {
             await openPrivateCloudProductEditSubmitModal({
               productData: formData,
-              originalProductData: methods.getValues(),
+              originalProductData: currentProduct,
             });
           })}
           autoComplete="off"
@@ -194,7 +197,7 @@ export default privateCloudProductEdit(({ session }) => {
         </form>
       </FormProvider>
 
-      <SiloAccordion className="my-4" product={snap.currentProduct} />
+      <SiloAccordion className="my-4" product={currentProduct} />
     </div>
   );
 });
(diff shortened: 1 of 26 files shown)


app/app/private-cloud/products/(product)/[licencePlate]/edit/page.tsx
```diff
@@ -87,6 +87,7 @@ export default privateCloudProductEdit(({ session }) => {
       )(values, context, options);
     },
     defaultValues: {
+      hasRepositories: null,
       repositories: [],
       isAgMinistry: false,
```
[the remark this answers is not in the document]
Comment (Kolezhanchik) [reply app/app/private-cloud/products/(product)/[licencePlate]/edit/page.tsx:94]: `null` is only the initial controlled form value before `currentProduct` is available. Once the product loads, the effect resets the form using the persisted `hasRepositories` value, with true derived from a non-empty repositories array for backward compatibility. Therefore, the edit form does reflect the current stored data without requiring the user to reselect `Yes` or `No`

app/validation-schemas/private-cloud.ts
```diff
@@ -166,7 +174,8 @@ export const privateCloudEditRequestBodySchema = _privateCloudEditRequestBodySch
   .refine(validateDistinctPOandTl, {
     message: 'The Project Owner and Primary Technical Lead must be different.',
     path: ['primaryTechnicalLeadId'],
-  });
+  })
+  .superRefine(validateRepositorySelection);
```
MonicaG: The code between lines 130-179 seem to be pretty identical except for the `.merge(privateCloudProductWebhookBodySchema)` which is done in one but not the other. 

Maybe make a note with @danilomeireles about refactoring this to remove the duplicate with the re-architecting work he is doing.

app/app/public-cloud/products/(product)/[licencePlate]/edit/page.tsx
Comment (MonicaG) [inline app/app/public-cloud/products/(product)/[licencePlate]/edit/page.tsx:1]: Not sure if it's possible, but the useEffect in this code and the one in the private-cloud page.tsx look to be the same.  Is it possible to refactor into shared code that can be used by both pages? If it's not a big refactor to do maybe do as part of this PR. Otherwise maybe make a note with @danilomeireles about including this as an area for refactor as part of the architectural review/update.

MonicaG (commented):
Hi Zhanna, please see my comments. Let me know what you think and I can approve if the refactors will be done at a later time.

app/app/public-cloud/products/(product)/[licencePlate]/edit/page.tsx
Comment (Kolezhanchik) [reply app/app/public-cloud/products/(product)/[licencePlate]/edit/page.tsx:1]: The effects are similar for repository initialization, but the Public Cloud form also handles networking and environment values. The Private Cloud form has a different schema and quota logic. A shared hook may add more complexity than this small duplication, so I would kept them separate in this PR and cleaned up the reset dependencies.

app/validation-schemas/shared.ts
```diff
@@ -42,57 +46,31 @@ export const deleteRequestRejectBodySchema = z.object({
...
-      {
-        message: 'Repository must be hosted on GitHub, Bitbucket, or GitLab',
-      },
-    )
-    .refine(
-      (value) => {
-        try {
-          const pathSegments = new URL(value).pathname.toLowerCase().split('/').filter(Boolean);
+  url: z.string().trim().min(1, 'Repository URL is required'),
+});
```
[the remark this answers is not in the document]
Kolezhanchik: that was requested by PO

app/validation-schemas/security-config.ts
```diff
@@ -1,13 +1,16 @@
...
-});
+export const securityConfigSchema = z
+  .object({
+    licencePlate: z.string(),
+    repositories: repositoriesSchema,
+    hasRepositories: hasRepositoriesSchema,
+    context: z.union([z.literal(ProjectContext.PRIVATE), z.literal(ProjectContext.PUBLIC)]),
+    clusterOrProvider: z.string().optional(),
+  })
+  .superRefine(validateRepositorySelection);
```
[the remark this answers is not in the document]
Kolezhanchik: it was requested to remove all of validation for the repository link for now

app/components/form/Repositories.tsx
```diff
@@ -75,7 +108,7 @@ export default function Repositories({
         </Button>
       )}
 
-      {disabled && fields.length === 0 && (
+      {disabled && fields.length === 0 && hasRepositories === true && (
```
MonicaG: I think this check is not working? When `fields.length === 0` I think the hasRepositories would be either false or null.

In order for the message to be displayed I think it needs to be updated to  one of:

1) `hasRepositories !== true` - display the message if hasRepositories is either no (false) or null (no answer)
2) `hasRepositories === false` - display the message if have responded "no"
3) `hasRepositories === null` - display the message if they haven't made a choice

MonicaG (changes requested):
HI Zhanna,

I've tested the migration script and the public and private cloud forms. 
I couldn't get one message to show up for me in my testing. I think it's because one of the if statement needs to be updated. I've left a comment below.

I know there was the requirement to remove the checks on the URL. However, it is currently possible to put scripting, html etc in the url field. It doesn't get rendered on the screen (no javascript pop ups). However, the field is in the db. So if any emails, reports etc use it, it could cause an issue if someone put javascript into it. The "url" field in the field also isn't really a url now. It's probably ok? But we should be aware that it may not be a url.  Example: `repositories: [ { url: 'a' }, {url: 'b'}]`. Maybe we want to rename it to something other than url? Or at least make a note for later incase we need to do something where we're expecting a URL but it isn't.


Other than that it looks good!

app/components/form/Repositories.tsx
Kolezhanchik: Updated the condition. Now the message is shown in `read-only` mode when the repositories array is empty, regardless of the `hasRepositories` value. Thank you!

MonicaG (approved):
Looks good!

PR #7900, opened 2026-08-17
feat(2951): add admin GPU quotas for Emerald and KLAB2

Kolezhanchik:


app/helpers/quota-change.ts
MonicaG, line 1: Optional suggestion: For the function setting the `isQuotaUpgrade` const, I think it could use the namespaceKeys and resourceKeys from the `app/constants/private-cloud.tx` file and loop through them to do the comparison. That way this code wouldn't need to be updated if a new resource or namespace is added.

Although this assumes that every resource needs to have this check. Which it currently does.

Maybe something like:
```suggestion
import { namespaceKeys, resourceKeys } from '@/constants';

export const isQuotaUpgrade = (oldval: ResourceRequestsEnv, newval: ResourceRequestsEnv) =>
  namespaceKeys.some((namespace) =>
    resourceKeys.some((resource) => {
      const oldRequest = oldval[namespace][resource] ?? 0;
      const newRequest = newval[namespace][resource] ?? 0;

      return oldRequest < newRequest;
    }),
  );
```

app/helpers/quota-change.ts
```diff
@@ -10,14 +10,36 @@ export const isQuotaUpgrade = (oldval: ResourceRequestsEnv, newval: ResourceRequ
...
+    (oldval.test.gpu ?? 0) < (newval.test.gpu ?? 0) ||
     oldval.production.cpu < newval.production.cpu ||
     oldval.production.memory < newval.production.memory ||
     oldval.production.storage < newval.production.storage ||
+    (oldval.production.gpu ?? 0) < (newval.production.gpu ?? 0) ||
     oldval.tools.cpu < newval.tools.cpu ||
     oldval.tools.memory < newval.tools.memory ||
-    oldval.tools.storage < newval.tools.storage
+    oldval.tools.storage < newval.tools.storage ||
+    (oldval.tools.gpu ?? 0) < (newval.tools.gpu ?? 0)
```
MonicaG: Are the `??` required as the data has been backfilled by the migration script?

app/helpers/quota-change.ts
```diff
@@ -10,14 +10,36 @@ export const isQuotaUpgrade = (oldval: ResourceRequestsEnv, newval: ResourceRequ
...
+  canManageGpu: boolean,
+): ResourceRequestsEnv {
+  const gpuEnabled = canManageGpu && (cluster === Cluster.EMERALD || cluster === Cluster.KLAB2);
+
+  return Object.fromEntries(
+    Object.entries(resourceRequests).map(([namespace, requests]) => [
+      namespace,
+      {
+        ...requests,
+        gpu: gpuEnabled ? requests.gpu ?? 0 : 0,
```
MonicaG: Is the `??` needed since the data is backfilled by the migration script?

app/helpers/auto-approval-check.ts
```diff
@@ -79,9 +86,18 @@ export async function getQuotaChangeStatus({
...
+    const gpuDiff = requestedGpu - currentGpu;
+
+    if (gpuDiff !== 0) {
+      hasChange = true;
+    }
 
-    if (hasSignificantIncrease) {
+    if (gpuDiff > 0) {
+      hasIncrease = true;
+      hasSignificantIncrease = true;
```
MonicaG, line 100: Just checking that any increase should set `hasSignificantIncrease` to true. It doesn't need a calculation like above?

app/components/private-cloud/sections/Quotas.tsx
```diff
@@ -196,7 +224,7 @@ export default function Quotas({
                       classNames={{ wrapper: 'mt-3' }}
                       options={{ valueAsNumber: true }}
                       min={0}
-                      max={resourceKey === 'cpu' ? 64 : resourceKey === 'memory' ? 128 : 512}
+                      max={resourceMaxValue[resourceKey]}
```
MonicaG, line 227: Nice refactoring job here

app/app/private-cloud/products/(product)/[licencePlate]/edit/page.tsx
Comment (MonicaG) [inline app/app/private-cloud/products/(product)/[licencePlate]/edit/page.tsx:1]: The 3 `page.tsx`, `create.ts`, `update.ts` and the `route.ts` files all have a similar line:
```
const canManageGpu = !!session?.isAdmin || !!session?.permissions.reviewAllPrivateCloudRequests;
```
Can this logic be moved into a common helper function?

app/helpers/quota-change.ts
MonicaG: Why is there a `??` check for `gpu` but not the other fields? If the migration script populates the `gpu` field is this check needed here? Or alternatively, should the other fields have the check too?

MonicaG (changes requested):

app/helpers/auto-approval-check.ts
Kolezhanchik: Yes, this is intentional. Any GPU quota increase should require manual review, so it is treated as a significant increase regardless of the amount. Unlike CPU/memory/storage, GPU doesn't go through the metrics-based auto-approval calculation.

app/helpers/quota-change.ts
Kolezhanchik: Good suggestion — resourceKeys and namespaceKeys are the source of truth here. I’ll refactor `isQuotaUpgrade` to iterate over them. thank you!

app/app/private-cloud/products/(product)/[licencePlate]/edit/page.tsx
Comment (Kolezhanchik) [reply app/app/private-cloud/products/(product)/[licencePlate]/edit/page.tsx:1]: good point. I'll extract the shared GPU permission check into a helper

app/helpers/quota-change.ts
Kolezhanchik: you are right, the migration backfills gpu for existing records, the ?? 0 fallback is redundant here. I’ll remove it.

MonicaG (approved):

PR #7954, opened 2026-08-25
chore(3022): move GitHub account fields to separate model

Kolezhanchik:


app/app/api/users/_operations/update-github.ts
```diff
@@ -89,26 +101,50 @@ export default async function updateGitHubOp({
...
-      githubUsername,
-      githubAccountId: validation.user.accountId,
-    },
-    select: {
-      id: true,
-      githubUsername: true,
-      githubAccountId: true,
-    },
-  });
+  let updatedUser;
```
MonicaG: why the change from `const` to `let` here?

wrnu (approved):
Overall, looks good.

One thing I did notice is that there is no data migration, but I'm not sure if this has made it to prod or not.

Perhaps a quick note on the PR that no migration is required if it has not

